PR #162, opened 2022-06-30
feat: guess gst category from gstin where API is not called

### LINKED ISSUE #121: set GST Category from GSTIN in address quick entry
Currently, this gets set from the API. But there is no fallback for users without API / if API call fails.

DaizyModi:
- For quick entry for party and address.
- Party and address doctypes.


india_compliance/gst_india/utils/gstin_info.py
```diff
@@ -90,6 +90,22 @@ def _extract_address_lines(address):
     return address_line1, address_line2
 
 
+@frappe.whitelist()
+def get_gst_category_from_gstin(gstin):
+    gst_category = "Unregistered"
+
```
vorasmit: If not gstin:
if country is not india:
gst_category is overseas

india_compliance/gst_india/utils/gstin_info.py
```diff
@@ -90,6 +90,22 @@ def _extract_address_lines(address):
...
+    if UNBODY.match(gstin):
+        gst_category = "UIN Holders"
+    elif TDS.match(gstin):
+        gst_category = "Tax Deductor"
+    elif OVERSEAS.match(gstin):
+        gst_category = "Overseas"
+    elif REGISTERED.match(gstin):
+        gst_category = "Registered Regular"
+
+    return frappe._dict(gst_category=gst_category)
```
vorasmit, line 106: no need to return object. you can just return gst_category

vorasmit (dismissed):

india_compliance/gst_india/utils/gstin_info.py
```diff
@@ -92,6 +93,31 @@ def _extract_address_lines(address):
...
+        return "Tax Deductor"
+
+    if REGISTERED.match(gstin):
+        return "Registered Regular"
+
+    if UNBODY.match(gstin):
+        return "UIN Holders"
+
+    if OVERSEAS.match(gstin):
+        return "Overseas"
```
sagarvora: this should happen in frontend itself - even if it means we need to duplicate some constants.

india_compliance/gst_india/client_scripts/address.js
```diff
@@ -5,6 +5,7 @@ const DOCTYPE = "Address";
 validate_gstin(DOCTYPE);
 update_gstin_in_other_documents(DOCTYPE);
 show_overseas_disabled_warning(DOCTYPE);
+set_gst_category(DOCTYPE);
```
sagarvora, line 8: Where is this defined?

india_compliance/public/js/utils.js
```diff
@@ -55,9 +56,26 @@ Object.assign(ic, {
         if (GSTIN_REGEX.test(gstin) && is_gstin_check_digit_valid(gstin)) {
             return gstin;
         }
-    }
-});
+    },
 
+    set_gst_category(frm) {
```
sagarvora: Let's call this `guess_gstin_category` with params `gstin` and `country`

Return value can be set into `gstin_category` field manually.

PR #177, opened 2022-07-04
fix: added hsn wise item list for more than 250 records

### LINKED ISSUE #90: Add support for HSN wise EWB
## Current Scenario and issue

- e-Waybill needs only HSN breakup of item details, but item-wise details are being sent to the portal.
- Nothing incorrect about this, but e-Waybill only supports 250 line items in e-Waybill.
- This is generally enough for 99.99% of transactions.
- If there are more than 250 items, its currently not supported by the API.

## Proposed Solution

- if there are more than 250 items, hsn-wise summary of items can be uploaded to e-Waybill.
- `get_item_list` from transaction data can be over-ridden in e_waybill utils, and HSN summary can be returned if there are more than 250 items.
- current validation of 250 items can then be removed.

DaizyModi:
Closes: #90

### DIFF
--- a/india_compliance/gst_india/utils/e_waybill.py
+++ b/india_compliance/gst_india/utils/e_waybill.py
@@ -467,6 +467,8 @@ class EWaybillData(GSTTransactionData):
 
         self.get_transaction_details()
         self.get_item_list()
+        if len(self.item_list) > 250:
+            self.get_hsn_wise_item_list()
         self.get_party_address_details()
 
         return self.get_transaction_data()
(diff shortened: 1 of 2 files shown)


vorasmit (dismissed):

india_compliance/gst_india/utils/e_waybill.py
```diff
@@ -469,7 +469,7 @@ def get_data(self):
         self.get_item_list()
         self.get_party_address_details()
 
-        return self.get_transaction_data()
+        return self.sanitize_data(self.get_transaction_data())
```
sagarvora: this is already being done conditionally inside `get_transaction_data`. Why are we doing it here?

PR #178, opened 2022-07-05
fix: disable accounts settings incompatible with app

### LINKED ISSUE #146: disable accounts settings incompatible with app
## Current Issue

Following are the accounts settings that need to be handled

image

- **Determine address tax category form:** This should be always Billing Address for India specific users. Inter State vs Intra State is based on Billing Address GSTIN, So this should be set to this by default.

- **Automatically add taxes and charges from item tax template:** This should be always untick by default. As per current design, we are having all taxes (input, output, reverse charge) in one template. If this is enabled, it will fetch all taxes and this is not what a user wants.

## Proposed Solution

- Post Install Patch for above fields.
- Property setter to make this fields Read Only.
- Description to make users aware that these settings are overridden.

image

DaizyModi:
**Changes**
- Set these two fields read-only by property setters for India Compliance App.
- Property setters for Tax settings Label and Collapsible
- Setup Default Accounts Settings

image

Closes: #146


india_compliance/gst_india/setup/property_setters.py
```diff
@@ -59,6 +59,30 @@ def get_property_setters():
...
+            "doctype": "Accounts Settings",
+            "fieldname": "add_taxes_from_item_tax_template",
+            "property": "read_only",
+            "value": "1",
+        },
+        {
+            "doctype": "Accounts Settings",
+            "fieldname": "tax_settings_section",
+            "property": "label",
+            "value": "Tax Settings (overridden by India Compliance)",
```
vorasmit:
```suggestion
            "value": "Tax Settings (Overridden by India Compliance)",
```

vorasmit (approved):

sagarvora:
Screenshot-2022-07-16-004931

PR #187, opened 2022-07-08
feat: restrict `India Compliance Account` page in case of ic-api-enabled

DaizyModi:


sagarvora:
Can't have monkey patches in this app, Daizy.

PR #75, opened 2022-04-21
feat: Purchase Reconciliation Tool

pruthvi145:
## TODO

 - [x] Download GSTR2A
 - [x] Download GSTR2B
 - [ ] Write Tests for GSTR2A and GSTR2B
 - [ ] Tool

PR #192, opened 2022-07-10
feat: Purchase Reconciliation Tool

vorasmit:
#75 continued here


india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.py
```diff
@@ -0,0 +1,694 @@
...
+
+def _get_first_day(month, year):
+    """Returns first day of the month"""
+    return datetime.strftime(datetime(year, month, 1), "%Y-%m-%d")
+
+
+def _get_last_day(month, year):
+    """Returns last day of the month"""
+    return datetime(year, month, monthrange(year, month)[1])
+
```
sagarvora: These functions need to be consistent in what they return.

india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.py
```diff
@@ -0,0 +1,694 @@
...
+    end_date = min(fiscal_year.end_date, _getdate(return_type))
+
+    # latest to oldest
+    return tuple(reversed(_get_periods(fiscal_year.start_date, end_date)))
+
+
+def _get_periods(start_date, end_date):
+    """Returns a list of month (formatted as `MMYYYY`) in given date range"""
+
+    return pd.date_range(start_date, end_date, freq="MS").strftime("%m%Y").tolist()
```
sagarvora: We need to remove `pandas` as a dependency (reasons: [1](https://github.com/frappe/erpnext/pull/30598#issuecomment-1090978603) [2](https://github.com/frappe/erpnext/issues/27047)). It is an overkill for this use-case.

sagarvora: I think this answer can achieve same result:
https://stackoverflow.com/a/28290050/4767738

```
import datetime
from dateutil.rrule import rrule, MONTHLY

strt_dt = datetime.date(2001,1,1)
end_dt = datetime.date(2005,6,1)

dates = [dt for dt in rrule(MONTHLY, dtstart=strt_dt, until=end_dt)]
```

india_compliance/gst_india/utils/gstr/gstr.py
```diff
@@ -0,0 +1,116 @@
...
+def get_mapped_value(value, map):
+    return map.get(value)
+
+
+class GSTR:
+    # Maps of API keys to doctype fields
+    KEY_MAPS = frappe._dict()
+
+    # Maps of API values to doctype values
+    VALUE_MAPS = frappe._dict(
```
sagarvora, line 18: Shouldn't be a class var if it is designed to be mutated. Class vars are shared across all instances.

vorasmit:
- [x] Search Index for link field in GST Inward Supply
- [x] Send as text to frontend (Frappe)
- [x] Don't pop list or lookup in the list
- [x] Flexibility for 2A download
- [x] Respect SEZ or Overseas in 2A or 2B Download
- [x] BOE in reco tool
- [ ] Create new invoice in new tab

vorasmit:
continued in #916

PR #196, opened 2022-07-11
test: `get_gstin_info`

vorasmit:


india_compliance/gst_india/utils/test_gstin_info.py
```diff
@@ -0,0 +1,123 @@
+import unittest
+from unittest.mock import Mock, patch
```
ankush, line 2: You might wanna try https://github.com/getsentry/responses for better http response mocking and request validating. 😄

vorasmit:
Thanks, will surely look into it. Would make things simpler.

PR #204, opened 2022-07-22
fix: Add HSN Codes from overseas invoices in HSN wise summary report

deepeshgarg007:
Port Of: https://github.com/frappe/erpnext/pull/31665

### DIFF
--- a/india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.py
+++ b/india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.py
@@ -34,7 +34,7 @@ def _execute(filters=None):
     added_item = []
     for d in item_list:
         if (d.parent, d.item_code) not in added_item:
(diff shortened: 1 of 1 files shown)


vorasmit:
## Issues with existing report

1. Stock Qty, Total amount and taxable amount are coming in multiple of no of rows for taxes and charges.
Solution: Query for taxes can be a sub-query joined to the grouped values of invoice. Similarly, HSN description can be a different sub-query directly applied to summary.

2. Taxable amount is not the true taxable amount and is fetch from `base_net_amount` from item. Instead can be fetch from `taxable_value`.

3. No need to show extra columns for charges, other than taxes in columns.

4. Stock UOM should be one of the standard UOMs allowed by GST Portal. Similar to constants > e_waybill.py

5. JSON export is not as required by GST Portal and can be improved.

6. Since this may require a new query to be written, best if written in frappe.qb.


## Screenshot of columns as per GST Offline Tool
image

## JSON Format
image

## Schema
image

## Current Summary vs Invoice for your reference
Compare for HSN 73041990. Values are showing in multiples of 4 as there are 4 taxes & charges rows.
image


image

vorasmit:
Closed with #209

PR #207, opened 2022-07-29
chore: remove redundant fetching for address fields

sagarvora:
See also: https://github.com/frappe/erpnext/pull/31733

This fetching isn't needed because it already happens in `erpnext.accounts.party.set_address_details`

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -2,7 +2,6 @@ import json
 
 import frappe
 from frappe import _, bold
-from frappe.model.utils import get_fetch_values
 from frappe.utils import cint, flt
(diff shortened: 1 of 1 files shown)


india_compliance/gst_india/overrides/transaction.py
```diff
@@ -479,19 +478,6 @@ def get_regional_round_off_accounts(company, account_list):
 
 
 def update_party_details(party_details, doctype, company):
```
vorasmit, line 480: Is this function needed?

sagarvora: in the current design, yes. all keys in `gst_details` don't get set in `party_details` as of now, so `party_details.update...` is reqd.

we could get rid of `gst_details` entirely, but there are good reasons to not do that. anyway, I don't think it's within this PR's scope..

PR #223, opened 2022-08-17
feat: simplified tax setup for india

vorasmit:
## TODOs
- [ ] Add documentation link for simplified tax setup
- [ ] Add documentation
- [ ] Hide Tax category (settings)

### Improved tax defaults
- Removed tax categories
- Added appropriate config to tax templates

### Improved custom fields
- Tax config fields being created in Tax Templates instead of Tax Categories
- Removed gst_state config from Tax Category as state-wise accounts are not supported
- Added field in supplier for a reverse charge so that it can be set to default (only for Registered Regular GST category)

### GST Settings
- Added configuration to have default reverse charge for purchases from unregistered suppliers (with threshold)
- Added patch such that it can be executed from the front-end (at the option of the user)

### Get Party Details for transactions
- Split get details for party details and tax template. This ensures fewer get_fectch for changes in GST fields.
- Set reverse charge as per config in the supplier when the supplier is changed (not used fetch from as its a check field -- cannot be overwritten).

### Other changes
- Post Install patch and patch for existing users
- Validate tax accounts for taxes and charges template

### Screenshots

image

image

image


india_compliance/patches/post_install/remove_tax_category.py
```diff
@@ -0,0 +1,153 @@
...
+            update_values = {"is_inter_state": category.is_inter_state}
+            if doctype == "Purchase Taxes and Charges Template":
+                update_values["is_reverse_charge"] = category.is_reverse_charge
+
+            doc.db_set(update_values)
+
+    if show_alert:
+        frappe.msgprint(
+            "State wise Tax Categories are no longer supported. If you maintain state wise GST accounts, please merge them manually."
+        )
```
sagarvora: cant `msgprint` in a patch. use `secho`. also why are we warning about the same thing twice? this should already be warned about in the earlier post install patch.

india_compliance/patches/post_install/remove_tax_category.py
```diff
@@ -0,0 +1,154 @@
...
+        click.secho(
+            "Looks like you are using State wise Tax Categories. They are no longer required and hence not supported.",
+            color="yellow",
+        )
+
+
+def opt_out_needed():
+    categories = frappe.get_all(DOCTYPE, pluck="name")
+    for category in categories:
+        if category not in DEFAULT_CATEGORIES:
```
deepeshgarg007: In-state and Out-state are just titles and there might be cases that users might be using the old config with tax category with some other non-standard title, there are chances such users might be affected by this patch

india_compliance/public/js/transaction.js
```diff
@@ -94,3 +157,38 @@ function validate_overseas_gst_category(doctype) {
...
+}
+
+function toggle_reverse_charge(frm, enable) {
+    if (
+        (enable && frm.doc.is_reverse_charge) ||
+        (!enable && !frm.doc.is_reverse_charge)
+    )
+        return;
+
+    console.log("here", enable);
```
deepeshgarg007: Missed a log statement

india_compliance/patches/post_install/remove_tax_category.py
vorasmit: This patch doesn't delete GST Category (as set in various docs) if configured in Customer, Supplier, Address, or Item Tax.
`opt_out_needed` makes a check for the same.

Users in such scenarios can manually trigger the delete (Switch to Simplified Tax Setup - Screenshot 1) from GST Settings.

Updated: Not making a check for Default categories, as there can be non-standard categories.

rmehta:
Is this still valid?

vorasmit:
We can reopen when we are ready for this.

amrit-citrusleaf:
@vorasmit - can we please reopen this PR? This will greatly simplify GST setup 🙌🏼

PR #237, opened 2022-09-06
feat: add GST Ledger and GST Balance report

### LINKED ISSUE #255: Single Input and Output GST accounts
In v14, we can only have single Input/Output/RCM CGST/SGST/IGST accounts defined under GST settings. However, a company may have multiple GST numbers. In this case we need to have separate GST ledgers to track input/output GSTIN-wise. How can we go about this?

vorasmit:


vorasmit:
continued: #748

PR #240, opened 2022-09-07
feat: delete e-Waybill on cancel

vorasmit:
On cancel of e-Waybill, delete any existing attachments for old e-Waybills as they are not needed/meant to be reused.
For historical references, logs are available and can be printed again if needed from there.

Possible TODOs:

* [x] Watermark e-Waybill as cancelled in print format (HTML).

image

This is similar to what govt does to ensure it is not misused.

### DIFF
--- a/india_compliance/gst_india/client_scripts/e_waybill_actions.js
+++ b/india_compliance/gst_india/client_scripts/e_waybill_actions.js
@@ -19,6 +19,7 @@ function setup_e_waybill_actions(doctype) {
                     message: __("e-Waybill PDF attached successfully"),
                 });
             });
+            frappe.realtime.on("reload_doc", () => frm.reload_doc());
         },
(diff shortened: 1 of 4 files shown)


sagarvora:
PDF doesn't support CSS grids. Not surprising:
image

india_compliance/gst_india/client_scripts/e_waybill_actions.js
```diff
@@ -19,6 +19,7 @@ function setup_e_waybill_actions(doctype) {
                     message: __("e-Waybill PDF attached successfully"),
                 });
             });
+            frappe.realtime.on("reload_doc", () => frm.reload_doc());
```
sagarvora, line 22: this is bad design. it is basically a socketIO event to make an API call.

PR #250, opened 2022-09-15
test: test case for `generate_e_invoice`

DaizyModi:


india_compliance/gst_india/utils/test_e_invoice.py
```diff
@@ -0,0 +1,230 @@
...
+            self.BASE_URL + "/test/ei/api/invoice",
+            body=json.dumps(self.MOCK_EWAYBILL_IRN_RESPONSE),
+            status=200,
+        )
+
+        responses.add(
+            responses.GET,
+            url=self.BASE_URL + "/test/ewb/ewayapi/getewaybill",
+            match=[matchers.query_param_matcher(params)],
+        )
```
vorasmit: Not needed.
If you generate an e-Invoice with transporter details. It should automatically generate e-Waybill without the need to generate it again.

india_compliance/gst_india/utils/test_e_invoice.py
```diff
@@ -0,0 +1,387 @@
...
+            re.compile(r"^(e-Invoice is not applicable for invoices before.*)$"),
+            validate_e_invoice_applicability,
+            si,
+        )
+
+        frappe.db.set_single_value(
+            "GST Settings", "e_invoice_applicable_from", get_datetime()
+        )
+
+    def _generate_e_invoice(self, doc, result):
```
vorasmit: Why are you duplicating these functions in test cases?

What is the objective of a test case if you are testing a test function?

DaizyModi: Yes, I just observe it yesterday. I will update it.

PR #256, opened 2022-09-26
test: test cases for validating  file `transaction_data.py`

DaizyModi:


india_compliance/gst_india/overrides/test_transaction_data.py
```diff
@@ -0,0 +1,197 @@
...
+    def setUpClass(cls):
+        frappe.db.savepoint("before_test_transaction_data")
+
+    @classmethod
+    def tearDownClass(cls):
+        frappe.db.rollback(save_point="before_test_transaction_data")
+
+    @classmethod
+    def setUp(cls):
+        pass
```
vorasmit: Not needed.

Also add item_tax_details

DaizyModi: item_tax_details is already called by function `get_all_item_details`. Added case for `get_address_details`

vorasmit (approved):

PR #271, opened 2022-10-07
fix: use lower version of Vue

DaizyModi:
**Before**

`bench build` breaks due to version dependency. Frappe upgrades Vue 2 to 3

**Fix**

1. Used required lower version of Vue for develop branch of IC.
2. Added version-14 branch Frappe and ERPNext in `server-tests.yml` file.


.github/workflows/server-tests.yml
```diff
@@ -93,6 +93,7 @@ jobs:
                   bash ${GITHUB_WORKSPACE}/.github/helper/install.sh
               env:
                   BRANCH: develop
+                  FRAPPE_BRANCH: version-14
```
vorasmit:
```suggestion
                  BRANCH: version-14
```

PR #327, opened 2022-11-09
feat: validate distance in ewaybill if to and from pincodes are same

### LINKED ISSUE #262: e-Waybill distance validation
Ensure distance cannot be zero where to and from Pincode are same.

image

DaizyModi:
Closes: #262


vorasmit:
@sagarvora

Distance in e-Waybill is just for determining the validity of e-Waybill and has no legal standing.
Hence, its accuracy is not of prime importance.

Most users go with zero distance for the same reason.

**Proposal:**

Let's set the distance to 1 Km (instead of throwing an error) where it's not set. If users are particular about it, they would anyway enter this manually.

Where validity is not sufficient, it can be extended.

This would mean less hassle and give better UX.

PR #332, opened 2022-11-10
fix: reduce min length required for address line 1

sagarvora:
@vorasmit kindly merge if okay.

Based on this spec:
https://einv-apisandbox.nic.in/version1.03/generate-irn.html#JSONSchema


vorasmit:
This is in contrast to the JSON preparation utility (Version 1.01).
https://einvoice1.gst.gov.in/Others/BulkGenerationTools

However, API seems to be on version 1.03 and this works.

vorasmit (approved):

PR #333, opened 2022-11-10
fix: hsn summary calculation

rtdany10:
**Problem:**
If you raise a GST invoice with the same item multiple times in the invoice with different UOMs, the report only shows data of 1 uom and skips the rest.

Also, if you add the same item twice in the invoice with different HSN Code(eg: Service items with same name and different HSN code), the report only shows data of either one.

**Cause:**
HSN wise Summary report grouped data in the SQL query by the following 4 parameters:
1. Parent, ie, Sales Invoice ID
2. HSN Code
3. Item Code
4. UOM

And in the `_execute` function, we grouped it by:
1. Parent
2. Item Code

[And then added only 1 matching data to the report.](https://github.com/resilient-tech/india-compliance/blob/next/india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.py#L43)

This caused issue as only 1 entry from the 1st group was considered by the report.

**Solution:**
This PR changed the grouping to:
SQL:
1. Parent
2. HSN Code
3. Item Code

`_execute` function:
1. Parent
2. HSN Code
3. Item Code

Also re-wrote the test case to accommodate this scenario.

### DIFF
--- a/india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.py
+++ b/india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.py
@@ -40,7 +40,7 @@ def _execute(filters=None):
     data = []
     added_item = []
     for d in item_list:
-        if (d.parent, d.item_code) not in added_item:
+        if (d.parent, d.gst_hsn_code, d.item_code) not in added_item:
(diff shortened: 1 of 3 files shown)


rtdany10:
@sagarvora @vorasmit please review :)

rtdany10:
@vorasmit :grimacing:

vorasmit (approved):
Developer Note:

UOM grouping is irrelevant as the report is based on Stock UOM.
One item can only have one stock UOM.

PR #373, opened 2022-12-06
fix: reverse_charge field validation if present (POS Invoice exception)

### LINKED ISSUE #372: POSInvoice is getting validated for 'is_reverse_charge'
POSInvoice is getting validated for 'is_reverse_charge' which should not be there as per the file:
 
 https://github.com/resilient-tech/india-compliance/blob/version-14/india_compliance/gst_india/constants/custom_fields.py#L49-L100

It is causing this error below in Point of Sale Page:

Screenshot 2022-12-06 164355

ERROR:

image

vorasmit:
Closes: #372 

Reference to custom field creation:
https://github.com/resilient-tech/india-compliance/blob/next/india_compliance/gst_india/constants/custom_fields.py#L581

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -185,7 +185,7 @@ def validate_gst_accounts(doc, is_sales_transaction=False):
                 ).format(idx)
             )
 
-        if doc.is_reverse_charge and (
+        if doc.get("is_reverse_charge") and (
             idx := _get_matched_idx(rows_to_validate, all_valid_accounts)
         ):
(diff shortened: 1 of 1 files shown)


sagarvora (changes requested):
I think there are more instances in the same function that will also require use of `.get`

vorasmit:
> I think there are more instances in the same function that will also require use of `.get`

Other instance is applicable for purchase. Only this one is applicable for Sales.

sagarvora (approved):

sagarvora:
@mergifyio backport version-14

PR #381, opened 2022-12-08
fix: Use country as a fallback for address lines

deepeshgarg007:
Since "Address Line 2" is not mandatory in the Address master many times users don't update this and e-Invoice generation leads to the below error.

Screenshot 2022-12-07 at 11 14 17 PM

Added this as a fallback for "Address Line 1" and "City" fields as well as these are open data fields and post sanitization they may not satisfy the required conditions

Using country since its a Link/Auto Complete field and there and there are fewer changes of mistakes in this field

Integration Request Dump

```
[...]

PR #384, opened 2022-12-10
fix: handle erros while sanitizing mandatory value

DaizyModi:
Continue: #381 

Error Screenshots:
image

image

Tradename, Location, and Address Line1 cannot be empty and It should be valid characters while generating e-Waybill or e-Invoice.

### DIFF
--- a/india_compliance/gst_india/utils/transaction_data.py
+++ b/india_compliance/gst_india/utils/transaction_data.py
@@ -310,22 +310,35 @@ class GSTTransactionData:
 
         self.check_missing_address_fields(address, validate_gstin)
 
+        error_context = {
+            "doctype": "Address",
+            "docname": address.name,
+        }
+
         return frappe._dict(
             {
                 "gstin": address.get("gstin") or "URP",
                 "state_number": address.gst_state_number,
                 "address_title": self.sanitize_value(
-                    {"address_title": address.address_title}, 2, throw=True
+                    address.address_title,
+                    regex=2,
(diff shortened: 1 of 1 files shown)


india_compliance/gst_india/utils/transaction_data.py
```diff
@@ -396,15 +403,50 @@ def sanitize_value(
...
+                _(
+                    "Unicode characters are not allowed. Please edit details in English only."
+                )
+            )
 
         if regex:
             value = re.sub(REGEX_MAP[regex], "", value)
 
+        if not value:
+            return _throw(_("{0} is mandatory.").format(key))
```
vorasmit: You are already checking this in 428. Duplication is not needed here.

india_compliance/gst_india/utils/transaction_data.py
```diff
@@ -396,15 +403,50 @@ def sanitize_value(
...
+
         if not truncate and len(value) > max_length:
-            return
+            return _throw(
+                _("Value must be at most {0} characters long.").format(max_length),
+            )
```
vorasmit: No need to throw here. Let it fail silently as it's not used in mandatory fields.

india_compliance/gst_india/utils/transaction_data.py
DaizyModi: This validation is for value after regex is applied.

vorasmit (approved):

PR #439, opened 2023-01-04
fix(quick entry): missing link in new address

rtdany10:
Before:
image

After:
image


vorasmit:
Wasn't able to push changes here.
Hence continued in #444

PR #446, opened 2023-01-10
fix: missing HSN for variant

rtdany10:
https://user-images.githubusercontent.com/52111700/211530496-ed3299ae-4004-406d-a491-0d1e72a94390.mp4

### DIFF
--- a/india_compliance/gst_india/setup/__init__.py
+++ b/india_compliance/gst_india/setup/__init__.py
@@ -230,6 +230,8 @@ def _get_custom_fields_map(*custom_fields_list):
 
 def remove_item_variant_gst_map():
     fields_to_remove = ["gst_hsn_code", "is_nil_exempt", "is_non_gst"]
-    rows = frappe.db.get_all("Variant Field", {"field_name": ["in", fields_to_remove]}, pluck="name")
+    rows = frappe.db.get_all(
+        "Variant Field", {"field_name": ["in", fields_to_remove]}, pluck="name"
+    )
     for d in rows:
         frappe.delete_doc("Variant Field", d)
--- a/india_compliance/gst_india/setup/__init__.py
+++ b/india_compliance/gst_india/setup/__init__.py
@@ -31,7 +31,7 @@ def after_install():
 def before_uninstall():
     delete_custom_fields()
     delete_property_setters()
-    remove_variant_hsn_map()
+    remove_item_variant_gst_map()
 
 
 def create_custom_fields():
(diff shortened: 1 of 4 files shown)


vorasmit:
Add this as a part of `install.py` also (Post Install Patch).
This will ensure that sites migrating from v13 can have this.

india_compliance/patches/post_install/add_hsn_to_item_variant_copy_fields.py
```diff
@@ -0,0 +1,11 @@
...
+
+"""
+This patch makes sure hsn code is copied over to variant item when created from template
+"""
+
+
+def execute():
+    frappe.get_doc("Item Variant Settings").append(
+        "fields", {"field_name": "gst_hsn_code"}
+    ).save()
```
vorasmit: Should be done only if it's not already there?

rtdany10: Have added a check :)
thanks!

india_compliance/install.py
```diff
@@ -10,6 +10,7 @@
 POST_INSTALL_PATCHES = (
     ## ERPNext
     "setup_custom_fields_for_gst",
+    "add_hsn_to_item_variant_copy_fields",
```
vorasmit: EPRNext Patches are old patches upto v13 kept for users upgrading from older versions to v14.

Add this patch under India Compliance (as commented)

india_compliance/gst_india/setup/__init__.py
```diff
@@ -225,3 +226,14 @@ def _get_custom_fields_map(*custom_fields_list):
...
+
+def remove_variant_hsn_map():
+    row = frappe.db.exists("Variant Field", {"field_name": "gst_hsn_code"})
+    if row:
+        ivs = frappe.get_doc("Item Variant Settings")
+        to_remove = [d for d in ivs.fields if d.field_name == "gst_hsn_code"]
+        ivs.fields.remove(to_remove[0])
+        for index, item in enumerate(ivs.fields):
+            item.idx = index
+        ivs.save()
```
vorasmit:
```suggestion
        frappe.delete_doc("Variant Field", row)
```

vorasmit:
This can also be done for other existing GST fields in Item.
- Is non gst
- Is Nil Exempt

india_compliance/gst_india/setup/__init__.py
rtdany10: I have changed the code to the suggested one, although it can mess up the idx on the table:
image

india_compliance/install.py
rtdany10: Have changed the ordering

PR #455, opened 2023-01-17
fix: Negative credit days error while generating e-Invoice

deepeshgarg007:
Many users apply payment terms on orders and not invoices. ERPNext has a provision to fetch the payment terms from the order and not recalculate them during invoicing again. In such cases, the payment due date can be before the invoice posting date. This leads to an error while generating e-Invoice for such invoices as negative credit days are calculated

image


vorasmit (approved):
There is conditional validation in ERPNext (to validate Due Date).
Ref: accounts_controller.py

However, we need to be sure here.

PR #458, opened 2023-01-19
fix: copy GST Item fields to its variants

vorasmit:
Continue: #446


sagarvora:
@mergifyio backport version-14-hotfix

Issue #288, opened 2022-10-10
Handle manual intervention for e-Waybills


vorasmit:
Case1.
Manually generated e-Waybill and updated.
Trying to cancel e-Waybill & e-Invoice.

Solution:
On cancel, if the e-Waybill log is not available. Try cancelling.

- [x]  e-Waybill No is not valid
   - Don't cancel the e-Invoice. Throw error.
- [x]  e-Waybill is already cancelled
   - Update e-Waybill is cancelled in the log is there and cancel e-Invoice
- [x]  e-Waybill not cancellable
   - Throw an error. Don't cancel the e-Invoice 
- [x]  Success
   - Update e-Waybill is cancelled in the log is there and cancel e-Invoice

PR #459, opened 2023-01-19
fix: error handling for manually cancelled e-waybill or e-Invoice

DaizyModi:
**Issue**

1. If e-Waybill is manually canceled from GSP Portal 
2. If the e-Invoice is manually canceled 
If a user tries to cancel it from India Compliance App which updates the `e-Waybill Log` and `e-Invoice Log`, it throws an error attached in the screenshot:

image

`e-Invoice` error message:

```
ValidationError: 9999 : Invoice is not active
```

**Fix**

Fix for issue: #288
1. Handle errors silently.
2. Allow users to cancel `e-Waybill` and Update the `e-Waybill Log` with cancel values.
3. Allow users to cancel `e-Invoice` and update the `e-Invoice Log` with cancel values.

cancelled_e_invoice


india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -115,31 +115,31 @@ def generate_e_invoice(docname, throw=True):
...
+    result = EInvoiceAPI(doc).cancel_irn(
+        e_invoice_data.get_data_for_cancellation(values)
+    )
 
     log_e_invoice(
         doc,
         {
-            "name": result.Irn,
+            "name": result.get("Irn")
+            or doc.irn,  # fallback to handle already cancelled IRN
```
vorasmit: always use doc.irn. This should be the same.

vorasmit:
LGTM: Tested.

Note
- For sandbox mode, cancel_ewaybill is called from `e_invoice.py` where such error handling is not done and would still throw error.

PR #469, opened 2023-02-01
fix: wrong item wise tax calculation

rtdany10:
Tax calculation goes wrong if the same tax account is repeated in the tax table.

https://github.com/frappe/ecommerce_integrations/blob/32a62fc199699af55077903eee14b09d935dff5a/ecommerce_integrations/shopify/order.py#L188-L211

Shopify orders, when synced, would have a manually calculated tax table, as seen in the above code.

### DIFF
--- a/india_compliance/gst_india/utils/transaction_data.py
+++ b/india_compliance/gst_india/utils/transaction_data.py
@@ -250,17 +250,15 @@ class GSTTransactionData:
                 continue
 
             tax_rate = self.rounded(item_tax[0], 3)
-            tax_amount = abs(self.rounded(item_tax[1]))
 
             # considers senarios where same item is there multiple times
-            if row.charge_type != "Actual":
-                tax_amount = abs(
-                    self.rounded(
-                        tax_rate * item.qty
-                        if row.charge_type == "On Item Quantity"
-                        else tax_rate * item.taxable_value / 100
-                    ),
-                )
+            tax_amount = abs(
+                self.rounded(
+                    tax_rate * item.qty
+                    if row.charge_type == "On Item Quantity"
+                    else tax_rate * item.taxable_value / 100
+                ),
+            )
 
             item_details.update(
                 {
--- a/india_compliance/gst_india/utils/transaction_data.py
+++ b/india_compliance/gst_india/utils/transaction_data.py
@@ -243,14 +243,14 @@ class GSTTransactionData:
 
             # Remove '_account' from 'cgst_account'
             tax = self.gst_accounts[row.account_head][:-8]
-            tax_rate = frappe.parse_json(row.item_wise_tax_detail).get(
+            item_tax = frappe.parse_json(row.item_wise_tax_detail).get(
                 item.item_code or item.item_name
             )
-            if not tax_rate:
+            if not item_tax:
                 continue
 
-            tax_rate = self.rounded(tax_rate[0], 3)
-            tax_amount = abs(self.rounded(tax_rate[1]))
+            tax_rate = self.rounded(item_tax[0], 3)
+            tax_amount = abs(self.rounded(item_tax[1]))
 
             # considers senarios where same item is there multiple times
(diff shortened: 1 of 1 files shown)


rtdany10:
@vorasmit please review

vorasmit:
Hi,

Thanks for this PR and bringing this to our notice.
For these two reasons this can cause further problems.

### Shopify Shipping charges:

https://github.com/frappe/ecommerce_integrations/blob/32a62fc199699af55077903eee14b09d935dff5a/ecommerce_integrations/shopify/order.py#L245-L284

Shipping charges here would give incorrect taxable value for your items. Its better to refactor this part of code, where shipping charges are added as Item to items table.

### Item wise tax details:

https://github.com/resilient-tech/india-compliance/pull/469/files#diff-fec96563b4b515be3c1c4180f8af308d86803d797120326e83080f24bb77f7aaR236

For e-Invoice or e-Waybill, its important to share item wise tax details. Your current implementation assumes that `Actual` means `Shopify` tax, which is not a good resumblence.

Also, this may give you incorrect results, where same item comes more than once in shopify order, as item_wise_tax_detail is based on `item_code`.

### Proposed Solution

Its better to club taxes in Shopify. What do you say?

rtdany10:
> Shipping charges here would give incorrect taxable value for your items. Its better to refactor this part of code, where shipping charges are added as Item to items table.

Not sure on how it will affect item's taxable value as the type is `Actual` and not `On Net Total`.

> Also, this may give you incorrect results, where same item comes more than once in shopify order, as item_wise_tax_detail is based on item_code.

Unlike ERPNext, same item can't be added as a separate row in Shopify, only quantity can be increased.

> Its better to club taxes in Shopify. What do you say?

Yes, this could tidy up things.
@ankush thoughts?

vorasmit:
> Not sure on how it will affect item's taxable value as the type is `Actual` and not `On Net Total`.

This is how taxable value is calculated (inherited from v13)
https://github.com/resilient-tech/india-compliance/blob/next/india_compliance/gst_india/overrides/transaction.py#L28-L61

It wouldn't consider any other charge in taxable value if `charge_type` is not `On Previous Row Total` for GST.

rtdany10:
@vorasmit I've opened a PR in ecom app to consolidate the tax table.

Can we also have this PR so that the iteration continues even if item is not found in the item wise split?

india_compliance/gst_india/utils/transaction_data.py
```diff
@@ -243,12 +243,13 @@ def update_item_tax_details(self, item_details, item):
...
-            tax_rate = self.rounded(
-                frappe.parse_json(row.item_wise_tax_detail).get(
-                    item.item_code or item.item_name
-                )[0],
-                3,
+            item_tax = frappe.parse_json(row.item_wise_tax_detail).get(
+                item.item_code or item.item_name
             )
+            if not item_tax:
+                continue
```
vorasmit, line 250: We are doing this only for GST Accounts.
Skipping a GST Account line would mean that incorrect taxes are calculated and would throw an incorrect error at a later stage as
` Items value + Taxes don't match Grand Total`

This is a required field at many places and this even gets set in the backend in ERPNext where it's created directly by API.

Can you think of a use case when this error could arise?

india_compliance/gst_india/utils/transaction_data.py
```diff
@@ -79,7 +79,7 @@ def update_transaction_tax_details(self):
                 continue
 
             tax = self.gst_accounts[row.account_head][:-8]
-            self.transaction_details[f"total_{tax}_amount"] = abs(
+            self.transaction_details[f"total_{tax}_amount"] += abs(
```
vorasmit, line 82: This would have worked correctly only for Shopify as `item_wise_tax_detail` was enforced with `dont_recompute_tax`.

Where it is auto-calculated, this would mean duplication of taxes.
eg:
image

PR #494, opened 2023-03-06
fix(transaction): only fetch party gst details if party name is available

rtdany10:
https://user-images.githubusercontent.com/52111700/223097236-3b14d76d-4185-42c3-b32b-b7171a214e6c.mp4


rtdany10:
@sagarvora @vorasmit please review

PR #518, opened 2023-03-14
perf: Optimization of HSN-wise summary report

deepeshgarg007:
- Make the description column optional and made a separate query to fetch the HSN Codes and avoid joins
- Removed unnecessary query in the for loop which was taking major chunk of the time in the report


india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.js
```diff
@@ -42,6 +42,11 @@ frappe.query_reports["HSN-wise-summary of outward supplies"] = {
 			"fieldtype": "Date",
 			"width": "80"
 		},
+		{
+			"fieldname":"show_description",
+			"label": __("Show HSN/SAC Description"),
+			"fieldtype": "Check"
+		},
```
sagarvora, line 49: forget to remove this @deepeshgarg007?

PR #531, opened 2023-03-28
fix: split e-invoice and e-waybill credentials

rtdany10:
Why? In multi-company scenario where one company requires e-Waybill and another requires both, the current setting will enable both for both companies. Splitting the services solves this issue.

image

ToDo: patch to split current credentials into two


rtdany10:
@sagarvora @vorasmit @deepeshgarg007 thoughts on this?

rtdany10:
@vorasmit

vorasmit:
continued in #611

PR #532, opened 2023-03-29
refactor: get_depreciation_amount according to frappe/erpnext#34646 [develop] (#532

anandbaburajan:
Refactoring `get_depreciation_amount`, which was refactored in https://github.com/frappe/erpnext/pull/34646.


sagarvora:
Thanks for the PR, @AnandBaburajan. Smit will review and merge it.

PR #533, opened 2023-03-29
refactor: get_depreciation_amount according to frappe/erpnext#34644

anandbaburajan:
Refactoring `get_depreciation_amount`, which was refactored in https://github.com/frappe/erpnext/pull/34644.


sagarvora:
@AnandBaburajan Pls. mark the PRs ready after they get merged in ERPNext, thanks.

PR #632, opened 2023-04-26
fix: allow same gstin in all transactions with validation

vorasmit:
This PR allows you to book a Sales Invoice to self.
https://discuss.frappe.io/t/billing-address-gstin-and-company-gstin-cannot-be-the-same/103864

Frappe Issue: ISS-23-24-00346

- Diluted validation to allow the same GSTIN. But cannot charge GST for the same.
- Added validations in e-Waybill for Sales Invoices and e-Invoice to skip such invoices.
- No similar validation is needed in GSTR-1 as it skips invoices without taxes.

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -142,6 +142,7 @@ def validate_gst_accounts(doc, is_sales_transaction=False):
     """
     Validate GST accounts
     - Only Valid Accounts should be allowed
+    - No GST account should be specified for transactions where Company GSTIN = Party GSTIN
     - If export is made without GST, then no GST account should be specified
     - SEZ / Inter-State supplies should not have CGST or SGST account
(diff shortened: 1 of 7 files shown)


india_compliance/gst_india/overrides/transaction.py
```diff
@@ -178,6 +179,19 @@ def _throw(message, title=None):
         doc.company, is_sales_transaction
     )
 
+    # Company GSTIN = Party GSTIN
+    party_gstin = doc.get("billing_address_gstin") or doc.get("supplier_gstin")
```
sagarvora, line 183: fieldnames aren't always reliable. someone may create a field called `billing_address_gstin` in a Purchase doc and the should not break code. use if condition => you already have `is_sales_transaction` available.

vorasmit: https://github.com/resilient-tech/india-compliance/pull/652/files

PR #656, opened 2023-05-08
feat: Created a Bill of Entry Summary Report

### LINKED ISSUE #599: Bill of Entry Improvement
### Is your feature request related to a problem? Please describe.

1. Creation of new Fields Bill Of Landing while creating of Bill Of Entry
2. In All Overseas transactions GST Amount would be calculated  Should be rounded off automatically
3. Required a proper Report with the below details. ( Can be Called Import Purchase Register )

PI No | PI Date | Amount | Bill of Entry | Bill Entry Date | Port Code | Assessable Value | Taxable Value | GST Rate | GST Amount | Custom Duty Paybale | Custom Duty Paid Entry No ( JV ) | Landed Cost Expense Account | Amount |

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

Shram007:
closes: #599


dineshpanchal93 (changes requested):
Can you please add here the Landed Cost details as well.

Shram007:
> Can you please add here the Landed Cost details as well.

- The current scenario of how Bill of Entry and Landed Cost Voucher are connected has many to one relations based on Purchase Invoice and Purchase Receipts which will make the report complicated. 
- In future, if better correlation is established then it can be included in the report.

vorasmit:
> Can you please add here the Landed Cost details as well.

Moreover, it's possible to have one-to-many relationship here, making it complex. Hence not covering it for now.

dineshpanchal93:
Getting the wrong IGST Value because the Report is fetched from the date of the Bill of Entry,

We get the IGST Input credit from the Purchase invoice Posting date.

vorasmit:
Hi,

As the name suggests, it's a bill of entry summary report. It would be not logical to show this by purchase invoice dates.
AFAIK, GST on the bill of entry can be claimed after its generated and paid, and not in the month of import.

eg: if you import in April, and the Bill of entry is in May, you should not be able to claim IGST in April. 

Hope this helps. Let me know if it's otherwise.

PR #665, opened 2023-05-08
fix: show zero rated items in GSTR-1 b2b and cdnr-reg report

### LINKED ISSUE #648: Zero Rated Bills in GST report not showing (B2B and Registered Credit Notes)

vorasmit:
closes: #648 

- Until now it was only accumulating this value for unregistered sales.
- Additionally, fix the condition for cdnr-reg
- Remove dependency on `item_tax_rate` as its not consistent and causes problems when there is no Item Tax Template

### DIFF
--- a/india_compliance/gst_india/report/gstr_1/gstr_1.py
+++ b/india_compliance/gst_india/report/gstr_1/gstr_1.py
@@ -494,21 +494,23 @@ class Gstr1Report(object):
 
                         for item_code, tax_amounts in item_wise_tax_detail.items():
                             tax_rate = tax_amounts[0]
-                            if tax_rate:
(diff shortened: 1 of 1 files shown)


rtdany10 (commented):
We will also require something like:
```
elif (
    not tax_rate
    and invoice in self.nil_exempt_non_gst
):
    taxable_value += abs(net_amount)
```
somewhere in lines between 286 and 308

vorasmit:
Fixed here already:

https://github.com/resilient-tech/india-compliance/pull/665/files#diff-7c7c19013fedfb0d2a15ea0c8bcf7ca0fcdf0d889e95b9fe117946e7adfc021fR497

Can you retry?

PR #691, opened 2023-05-17
feat: link to `Bill of Entry` in doctype dashboard

DaizyModi:
**Changes**
1. `Bill of Entry` link to Purchase Invoice dashboard
2. Added GST Details related to *Bill of Entry* in `GST Purchase Register` and `GST Itemised Purchase Register` Reports.

image


vorasmit:
extension to reports is not intuitive for end users and hence we are not pursuing this.

PR #697, opened 2023-05-18
fix: progressive item tax in credit note

rtdany10:
image

image


rtdany10:
Changes made:
As `response` is absolute value, took absolute value of `amount` before performing arithmetic operations.

rtdany10:
@vorasmit @sagarvora @DaizyModi

PR #701, opened 2023-05-19
fix(UX): Improve HSN code field

ankush:
1. Add description to let user know that HSN code can be filtered by descriptions.
2. Set property setter for `reqd` based on GST settings. This is required so it can be validated client side too.

### DIFF
--- a/india_compliance/gst_india/client_scripts/item.js
+++ b/india_compliance/gst_india/client_scripts/item.js
@@ -1,8 +1,20 @@
 frappe.ui.form.on('Item', {
+	onload: function(frm) {
+		if (!gst_settings.validate_hsn_code) return;
+		frm.set_query('gst_hsn_code', function() {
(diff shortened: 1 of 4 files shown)


vorasmit (approved):

ankush:
@vorasmit AFAIK mandatory depends on won't work in quick entry because full form doesn't load. 

Is there any problem with property setters?

vorasmit:
Oh, didn't know that.

However, its conditionally mandatory.
https://github.com/resilient-tech/india-compliance/blob/e4f1c234a1066bfc9552ccc152359ebdf9953727/india_compliance/gst_india/overrides/item.py#L12

vorasmit:
Hey,

I tested it locally on develop.
`mandatory_depends_on` is respected in the quick entry form.

However, `set_query` is not.

image

PR #752, opened 2023-06-07
fix: patch to update currency in tax accounts

ljain112:
Tax Accounts Currency is not set on creation in ERPNext


vorasmit:
Not required. Fixed in filters here:
https://github.com/frappe/erpnext/pull/35626

Issue #732, opened 2023-05-30
Possible API enhancements


vorasmit:
e-Invoice API Changes

- [x] Export details can be added to the e-Invoice. Country code is now mandatory

Better Bulk Actions

- [ ] bulk cancel e-Invoice or e-Waybill
- [x] bulk update transporter details
- [x] bulk generate e-Waybill
- [ ] bulk update GSTIN details for invalid GSTIN

Additional new APIs

- [x] Get IRN details by doc
- [x] Get e-Waybill details by IRN
- [ ] Get Rejected IRN details
- [x] #1038

PR #758, opened 2023-06-09
fix: Added export Details with country code in e-Invoice

DaizyModi:
Partial fix for #732


india_compliance/gst_india/utils/test_e_invoice.py
```diff
@@ -74,6 +74,41 @@ def test_request_data_for_different_shipping_dispatch_address(self):
...
+        si.update(
+            {
+                "gst_category": "Overseas",
+                "shipping_bill_number": "1234",
+                "shipping_bill_date": frappe.utils.today(),
+                "port_code": "INABG1",
+            }
+        )
+
+        if si.customer_address != si.port_address:
```
vorasmit: You should have complete control over the sales invoice that you create and should know if they are equal or not. Better --> Set a port address in the sales invoice.

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -654,3 +657,30 @@ def get_item_data(self, item_details):
                 "ExpDt": item_details.batch_expiry_date,
             },
         }
+
+    def get_export_details(self):
+        country_code = (
```
vorasmit: Directly from Billing Address only.
Throw if empty. (Add a link to country doctype and e-Invoice Master list)
Throw if the length is not equal to 2. (Add a link to country doctype and e-Invoice Master list)

india_compliance/gst_india/utils/transaction_data.py
```diff
@@ -409,6 +409,9 @@ def get_address_details(self, address_name, validate_gstin=False):
                     **error_context,
                 ),
                 "pincode": int(address.pincode),
+                "country_code": self.sanitize_value(
+                    get_country_code(address.country), min_length=2, max_length=2
```
vorasmit: Set only if not India. No need to sanitize.

india_compliance/gst_india/client_scripts/sales_invoice.js
```diff
@@ -26,5 +26,10 @@ frappe.ui.form.on(DOCTYPE, {
 
     before_submit(frm) {
         frm.doc._submitted_from_ui = 1;
-    }
+    },
+
+    refresh(frm) {
+        frm.set_df_property("port_code", "ignore_validation", 1);
```
vorasmit: Did you try onload?
Also, `set_df_property` can be part of set_port_code_options

india_compliance/public/js/utils.js
```diff
@@ -84,6 +86,21 @@ Object.assign(india_compliance, {
         if (UNBODY_REGEX.test(gstin)) return "UIN Holders";
         if (OVERSEAS_REGEX.test(gstin)) return "Overseas";
     },
+
+    set_port_code_options(frm) {
```
vorasmit: Add these to transaction.js
Utils should ideally not take `frm` as input.

india_compliance/gst_india/utils/test_e_invoice.py
DaizyModi: Already setting port address from test e_invoice records.

india_compliance/gst_india/utils/transaction_data.py
DaizyModi: Used `sanitize_value` for min and max length validations.

PR #795, opened 2023-06-24
fix: State Number of Dadra and Nagar Haveli & Daman and Diu has been updated

jay-parikh:
When creating new Address, if GST State is 'Dadra and Nagar Haveli' selected and State/Province is added as 'Dadra and Nagar Haveli' then getting error - "Please select a valid State from available options"


vorasmit:
Can you please provide evidence that there has been a split and that states have been updated? It's beneficial to have legal circulars instead of links to other websites.

jay-parikh:
> Can you please provide evidence that there has been a split and that states have been updated? It's beneficial to have legal circulars instead of links to other websites.

I didn't realized that DADRA AND NAGAR HAVELI AND DAMAN AND DIU (NEWLY MERGED UT). The GST state code for erstwhile UT of Daman and Diu was 25, prior to 26th January 2020.

But If you create new Address and check GST State field, somehow it's showing two separate state as shown in below image. 
Which is throwing an error - Invalid State Please select a valid State from available options
image

Frappe Framework: v14.40.0 (version-14)
India Compliance: v14.10.2 (version-14)
ERPNext: v14.27.12 (version-14)

sagarvora:
It seems like you may have exported customisations using the **Export Customisations** feature of Frappe, which causes a bug where changes made by India Compliance get overwritten by older exported customisations. Read more:

https://github.com/frappe/frappe/issues/21358


A temporary fix might be to[ sync custom fields](https://github.com/resilient-tech/india-compliance/blob/e9077f92835590d06b7ed9c9e90d8d3530bed7c6/india_compliance/gst_india/setup/__init__.py#L37) again, and export customisations again to ensure that they don't get overwritten during DB migration. Or even better, use and alternate approach instead of exporting customisations.

PR #800, opened 2023-06-27
fix: set e-invoice status as Failed on exception

vishakhdesai:
Set E-Invoice status as 'Failed" on exception in generate_e_invoice function

### DIFF
--- a/india_compliance/gst_india/utils/e_invoice.py
+++ b/india_compliance/gst_india/utils/e_invoice.py
@@ -130,6 +130,15 @@ def generate_e_invoice(docname, throw=True):
 
         return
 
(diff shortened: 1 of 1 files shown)


india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -121,6 +121,13 @@ def generate_e_invoice(docname, throw=True):
             _("Warning"),
             indicator="yellow",
         )
+
+        doc.db_set(
+            {
+                "einvoice_status": "Failed",
```
vorasmit: This will only work for validation errors. How will other cases be handled?

PR #805, opened 2023-06-29
fix(ux): auto retry generate e-invoice if failed because of Gateway Timeout

### LINKED ISSUE #699: enhance autogenerate: retry automatically if failing because of gateway timeout
- [x] if e-Invoice or e-Waybill failed because of gateway timeout, automatically retry every hr to generate e-invoice. maybe the e-Invoice status can be set to auto-retry.
- [x] Improved error for gateway timeout

photo_2023-05-18_15-56-39

vishakhdesai:
* Settings to Enable Retry e-Invoice Generation
* Scheduled cron job to auto retry e-Invoice generation based on settings
* Force e-Invoice generation on manual trigger
* e-Invoice Status will be set as **Auto-Retry** in case of `Gateway Timeout Error`

Closes: #699


india_compliance/gst_india/doctype/gst_settings/gst_settings.json
```diff
@@ -298,12 +300,24 @@
...
+   "fieldname": "e_invoice_auto_retry_after",
+   "fieldtype": "Datetime",
+   "label": "e-Invoice Auto Retry After"
+  },
+  {
+   "default": "60",
+   "description": "Ensures valid GSTIN in transactions by refreshing status periodically.",
+   "fieldname": "gstin_status_refresh_interval",
+   "fieldtype": "Int",
+   "label": "GSTIN Status Refresh Interval"
```
sagarvora: why?

india_compliance/gst_india/api_classes/base.py
```diff
@@ -213,5 +217,8 @@ def handle_http_code(self, status_code, response_json):
                 title=_("Invalid API Key"),
             )
 
+        if status_code == 504:
+            raise GatewayTimeoutError("Gateway Timeout")
+
```
sagarvora: what extra information is your error message giving you? are you showing a better error message than earlier?

india_compliance/gst_india/api_classes/base.py
```diff
@@ -12,6 +12,10 @@
 BASE_URL = "https://asp.resilient.tech"
 
 
+class GatewayTimeoutError(Exception):
+    pass
```
sagarvora: this is not the best place to define an exception, look at `frappe.exceptions` & `erpnext.exceptions`.

sagarvora: consider specifying default error message, like so: https://stackoverflow.com/questions/49224770/default-message-in-custom-exception-python

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -108,6 +128,24 @@ def generate_e_invoice(docname, throw=True):
...
+        doc.db_set({"einvoice_status": "Auto-Retry"})
+        e_invoice_auto_retry_after = now_datetime() + timedelta(minutes=5)
+
+        frappe.db.set_single_value(
+            "GST Settings", "e_invoice_auto_retry_after", e_invoice_auto_retry_after
+        )
+
+        frappe.msgprint(
+            _(
+                "Government services are currently experiencing downtime, resulting in high response times and a Gateway Timeout error. We apologize for the inconvenience caused. Your e-invoice generation will be automatically retried after {0} minutes."
```
sagarvora: This is technically self-contradictory. You can't be both down and have a higher response time.

india_compliance/gst_india/doctype/gst_settings/gst_settings.json
```diff
@@ -298,12 +300,24 @@
    "fieldname": "archive_party_info_days",
    "fieldtype": "Int",
    "label": "Archive Party Information (In Days)"
+  },
+  {
+   "fieldname": "e_invoice_auto_retry_after",
+   "fieldtype": "Datetime",
+   "label": "e-Invoice Auto Retry After"
```
sagarvora: do you really want user to be able to edit this?

india_compliance/hooks.py
```diff
@@ -238,6 +238,12 @@
     "Bill of Entry",
 ]
 
+scheduler_events = {
+    "all": [
```
sagarvora: note: this is is supposed to trigger every 4 minutes, but it may trigger as often as every 1 minute, based on user configuration. if more control is required, we can switch to cron expression

sagarvora (changes requested):
- `einvoice_status` custom field will need to be updated for new status. 

---

I vehemently disagree with forced auto-retry. Auto-retrying without configuration is okay when:
- it is being done after a small duration, like a second
- it is being done in the same process / request as the current one

In cases where you're making API requests in another process, it has to be configurable.

Proposal:
- Setting to "Automatically Retry e-Invoice Generation in case of API Timeout Errors".
- Can be enabled by default  + patch for existing sites?
- On disable, scheduled job type should be disabled by setting `stopped` as `1` in that document. Raise error as usual in that case.

india_compliance/gst_india/doctype/gst_settings/gst_settings.json
sagarvora: ideally, field should be hidden. it could even ideally be a "global". If it is being set in GST Settings, `update_modified` needs to be False in order to prevent unnecessary `TimeStampMismatchError`s

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -108,6 +128,24 @@ def generate_e_invoice(docname, throw=True):
...
+            ).format(str(e_invoice_auto_retry_after.replace(microsecond=0))),
+            _("Warning"),
+            indicator="yellow",
+        )
+
+        return
```
sagarvora: This should be `return False` instead. And then that should be captured by generate_e_invoices to prevent further execution. e-Invoice status for pending invoices should ideally be set in one query itself.

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -312,6 +359,34 @@ def validate_if_e_invoice_can_be_cancelled(doc):
         )
 
 
+def auto_retry_generate_e_invoice():
```
sagarvora:
```suggestion
def retry_e_invoice_generation():
```

india_compliance/gst_india/doctype/gst_settings/gst_settings.json
```diff
@@ -298,12 +300,24 @@
    "fieldname": "archive_party_info_days",
    "fieldtype": "Int",
    "label": "Archive Party Information (In Days)"
+  },
+  {
+   "fieldname": "e_invoice_auto_retry_after",
```
sagarvora: `retry_e_invoice_generation_after`

india_compliance/gst_india/utils/e_invoice.py
vishakhdesai: Is the following message appropriate?
"Government services are currently experiencing downtime, resulting in a Gateway Timeout error. We apologize for the inconvenience caused. Your e-invoice generation will be automatically retried after {0} minutes."

india_compliance/exceptions.py
```diff
@@ -0,0 +1,4 @@
+class GatewayTimeoutError(Exception):
+    def __init__(self, message="Gateway Timeout Error", *args, **kwargs):
```
vorasmit: Better error message

india_compliance/gst_india/doctype/gst_settings/gst_settings.json
```diff
@@ -314,7 +316,19 @@
...
-  }
+  },
+  {
+    "fieldname": "retry_e_invoice_generation_after",
+    "fieldtype": "Datetime",
+    "hidden": 1,
+    "label": "Retry e-Invoice generation after"
+   },
+   {
+    "default": "1",
```
vorasmit: Field position should ideally be in e-Invoice Settings, display depends on, etc

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -319,6 +363,57 @@ def validate_if_e_invoice_can_be_cancelled(doc):
...
+        )
+    )
+    if (
+        retry_e_invoice_generation_after
+        and retry_e_invoice_generation_after > now_datetime()
+    ):
+        return
+
+    queued_sales_invoices = frappe.db.get_all(
+        "Sales Invoice", filters={"einvoice_status": "Auto-Retry"}, pluck="name"
```
vorasmit, line 382: Add status to einvoice_status field in custom Fields.

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -108,6 +131,27 @@ def generate_e_invoice(docname, throw=True):
...
+        frappe.db.set_single_value(
+            "GST Settings",
+            "retry_e_invoice_generation_after",
+            retry_e_invoice_generation_after,
+            update_modified=False,
+        )
+
+        frappe.msgprint(
+            _(
+                "Government services are currently experiencing downtime, resulting in a Gateway Timeout error. We apologize for the inconvenience caused. Your e-invoice generation will be automatically retried after {0} minutes."
```
vorasmit: are currently slow (instead of experiencing downtime)

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -108,6 +131,27 @@ def generate_e_invoice(docname, throw=True):
...
+            ).format(str(retry_e_invoice_generation_after.replace(microsecond=0))),
+            _("Warning"),
+            indicator="yellow",
+        )
+
+        return False
```
vorasmit: Instead, raise and handle it above.

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -67,16 +69,25 @@ def enqueue_bulk_e_invoice_generation(docnames):
...
 
     for docname in docnames:
         try:
-            generate_e_invoice(docname)
+            generate_e_invoice(docname, throw=False, retry=retry)
+
+        except GatewayTimeoutError:
+            frappe.db.set_value(
+                "Sales Invoice",
+                {"name": ("in", docnames)},
```
vorasmit: irn filter

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -107,6 +123,31 @@ def generate_e_invoice(docname, throw=True):
             # IRN details cannot be provided as it is generated more than 2 days ago
             result = result.Desc if response.error_code == "2283" else response
 
+    except GatewayTimeoutError as e:
+        doc.db_set({"einvoice_status": "Auto-Retry"}, commit=True)
```
vorasmit: Incorrect status for settings not enabled

india_compliance/gst_india/doctype/gst_settings/gst_settings.py
```diff
@@ -49,10 +50,20 @@ def update_dependant_fields(self):
...
         # clear session boot cache
         frappe.cache.delete_keys("bootinfo")
 
+    def update_retry_e_invoice_scheduled_job(self):
+        if not self.has_value_changed("enable_retry_e_invoice_generation"):
+            return
+
+        frappe.db.set_value(
+            "Scheduled Job Type",
+            "e_invoice.enable_retry_e_invoice_generation",
```
vorasmit: .

PR #807, opened 2023-06-30
feat: provision to extend validity of e-Waybill

DaizyModi:


india_compliance/gst_india/client_scripts/e_waybill_actions.js
```diff
@@ -83,6 +83,14 @@ function setup_e_waybill_actions(doctype) {
                     () => show_update_transporter_dialog(frm),
                     "e-Waybill"
                 );
+
+                if (!frm.doc.gst_transporter_id && can_extend_e_waybill(frm)) {
```
vorasmit: .

vorasmit:
- [x] Handle is the latest data
- [x] No fields required. Instead, add a comment. However, update valid_upto so that it can be again extended if required.
- [x] GST transporter ID != Company ID

india_compliance/gst_india/utils/e_waybill.py
```diff
@@ -295,6 +299,59 @@ def update_transporter(*, doctype, docname, values):
...
 
 
+@frappe.whitelist()
+def extend_validity(*, doctype, docname, values):
+    doc = load_doc(doctype, docname, "submit")
+    values = frappe.parse_json(values)
+
+    doc.db_set(
+        {
+            "vehicle_no": values.vehicle_no.replace(" ", ""),
```
vorasmit, line 331: update only vehicle no

PR #812, opened 2023-06-30
refactor: separating lending doctypes from audit_trail_doctypes

anandbaburajan:
Loan management module is not a part of ERPNext anymore and has been separated into the new lending app. Hence separating the lending doctypes from audit_trail_doctypes, and only using them if the lending app is installed.

More: https://github.com/frappe/erpnext/pull/35522.

### DIFF
--- a/india_compliance/audit_trail/utils.py
+++ b/india_compliance/audit_trail/utils.py
@@ -7,7 +7,15 @@ def is_audit_trail_enabled():
 
 
 def get_audit_trail_doctypes():
-    return set(frappe.get_hooks("audit_trail_doctypes"))
+    audit_trail_doctypes = []
+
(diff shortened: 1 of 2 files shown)


sagarvora:
@anandbaburajan What are your thoughts on specifying `audit_trail_doctypes` hook directly in the Lending app instead of this?

We won't need separate hooks in that case, and it'd be much easier to manage if a new doctype which needs audit trail is added to the lending app.

I am removing the lending-related doctypes from the hooks in another commit: https://github.com/resilient-tech/india-compliance/commit/a1be014016a464396afeb4f17a76c96ed4f54177

sagarvora:
Closing in favor of https://github.com/frappe/lending/pull/3. @anandbaburajan please comment if you wish to reopen - with reasoning for the same.

PR #844, opened 2023-07-17
fix: update taxable value for non-gst_items

### LINKED ISSUE #838: Incorrect taxable value in item row
### Information about bug

The taxable value in GSTR-1 was not matching with the Invoice's Net Total. The reason for this is the incorrect taxable value in the item rows.

This only happens in two specific case
- The invoice is duplicated and then a normal item is replaced with the Non GST/ Nil Rated Item
- A regular item is selected and then replaced with a Non GST/Nil Rate Item

Taxable Value should be recalculated as 0 on server side even for such items

 

### Version

ERPNext Version - v14.28.1
Frappe Version - v14.40.1 
India Compliance Version - v14.10.3

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ljain112:
Closes: #838

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -83,6 +83,14 @@ def update_taxable_values(doc, valid_accounts):
         item.taxable_value += total_charges - apportioned_charges
 
 
+def update_taxable_values_for_non_gst(doc):
+    if doc.doctype not in DOCTYPES_WITH_TAXABLE_VALUE:
+        return
+
+    for item in doc.items:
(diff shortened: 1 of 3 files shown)


india_compliance/gst_india/overrides/transaction.py
```diff
@@ -83,6 +83,11 @@ def update_taxable_values(doc, valid_accounts):
         item.taxable_value += total_charges - apportioned_charges
 
 
+def update_taxable_values_for_non_gst(doc):
```
vorasmit, line 86: how about using no_copy?

PR #859, opened 2023-07-20
fix: patch to check version compatibility for dependencies on other apps

vorasmit:


india_compliance/patches/post_install/check_version_compatibility.py
```diff
@@ -0,0 +1,32 @@
...
+    # Stable Versions
+    "15.0.0": {"frappe": "15.0.0", "erpnext": "15.0.0"},
+    "14.12.0": {"erpnext": "14.29.0"},
+    "14.10.4": {"erpnext": "14.29.0"},
+    "14.0.0": {"frappe": "14.0.0", "erpnext": "14.0.0"},
+}
+
+
+def execute():
+    ic_version = frappe.get_attr("india_compliance.__version__")
```
sagarvora: `current_version`. And why are you using `frappe.get_attr`? that is so unnecessary when you can just import.

india_compliance/patches/post_install/check_version_compatibility.py
```diff
@@ -0,0 +1,32 @@
...
+
+    for _version, dependencies in VERSION_COMPATIBILITY.items():
+        if version.parse(ic_version) < version.parse(_version):
```
sagarvora: why parsing current version in loop unnecessarily?

india_compliance/patches/post_install/check_version_compatibility.py
```diff
@@ -0,0 +1,32 @@
...
+            continue
+
+        break
```
sagarvora: I think frappe `14.0.0` should still be checked even when checking erpnext `14.29.0`

india_compliance/patches/post_install/check_version_compatibility.py
```diff
@@ -0,0 +1,32 @@
...
+import frappe
+
+# Note: India Compliance versions in DECENDING order only
+VERSION_COMPATIBILITY = {
+    # Development Versions
+    "15.0.0-dev": {"frappe": "15.0.0-dev", "erpnext": "15.0.0-dev"},
+    # Stable Versions
+    "15.0.0": {"frappe": "15.0.0", "erpnext": "15.0.0"},
+    "14.12.0": {"erpnext": "14.29.0"},
+    "14.10.4": {"erpnext": "14.29.0"},
```
sagarvora: could just be one?

india_compliance/patches/post_install/check_version_compatibility.py
```diff
@@ -0,0 +1,32 @@
...
+    ic_version = frappe.get_attr("india_compliance.__version__")
+
+    for _version, dependencies in VERSION_COMPATIBILITY.items():
+        if version.parse(ic_version) < version.parse(_version):
+            continue
+
+        break
+
+    for app, required_version in dependencies.items():
+        app_version = version.parse(frappe.get_attr(f"{app}.__version__"))
```
sagarvora: please import. `frappe.get_attr` isn't the most pythonic for this purpose.

sagarvora:
```py
import frappe
import erpnext

...

current_versions = {"frappe": frappe.__version__, ...}

# use current versions
```

sagarvora: Note: we can also use `app.hooks.app_version` alternatively if internal variable needs to be avoided

india_compliance/patches.txt
```diff
@@ -1,5 +1,5 @@
 [pre_model_sync]
-india_compliance.patches.v15.check_version_compatibility
+india_compliance.patches.post_install.check_version_compatibility
```
sagarvora: perhaps patch should go entirely & we should check this before all migration? it doesn't seem expensive to me.

sagarvora (changes requested):

PR #900, opened 2023-07-27
fix: throw conditionally for e-Invoice validity date

### LINKED ISSUE #899: [Multi-company] Cannot invoice from companies which doesn't have e-invoice
https://github.com/resilient-tech/india-compliance/blob/8ed3bc2b1a065db5caa6f92e4d9ed350fe37daa2/india_compliance/gst_india/utils/e_invoice.py#L243-L302

Line 274, function `validate_e_invoice_applicability_date` doesn't use _throw but rather frappe.throw() which raises an error on invoicing from a company that doesn't have e-invoice enabled.

vorasmit:
closes: #899


rtdany10:
@vorasmit please initiate a patch release

PR #918, opened 2023-08-02
fix: patch update_company_fixtures fails for hierarchical multi-company setup

akurungadam:
`frappe.exceptions.ValidationError: Please add the account to root level Company`

basically, the new tax accounts should be added at the root level if the Companies are created as a tree


vorasmit:
closed with #919 . Created on dev branch.

PR #953, opened 2023-08-10
refactor: get_updated_rate_of_depreciation_for_wdv_and_dd instead of get_depreciation_amount

anandbaburajan:
In https://github.com/frappe/erpnext/pull/36587, I created a new `get_updated_rate_of_depreciation_for_wdv_and_dd` function, so now IC can just show the message and update the rate, and rest can happen in erpnext.


vorasmit:
- [ ] Update dependency after relevant PR in ERPNext is released

anandbaburajan:
> Update dependency after relevant PR in ERPNext is released

@vorasmit, where do I update the dependency? It's released in v14.35.0.

PR #1004, opened 2023-08-25
fix(ux): hsn query in item quick entry

### LINKED ISSUE #999: UX: Only show valid HSN code according to settings
Problem: 
- User opens quick entry form
- Picks HS code that's short
- validation error > full form opens. This is kinda bad UX. 

If Only >=6 character HSN are allowed then override search query to only return those.

vorasmit:
image

closes: #999

### DIFF
--- a/india_compliance/public/js/utils.js
+++ b/india_compliance/public/js/utils.js
@@ -128,6 +128,18 @@ Object.assign(india_compliance, {
         if (UNBODY_REGEX.test(gstin)) return "UIN Holders";
         if (OVERSEAS_REGEX.test(gstin)) return "Overseas";
     },
+
+    set_hsn_code_query(field) {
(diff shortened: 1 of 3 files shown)


india_compliance/public/js/utils.js
```diff
@@ -128,6 +128,18 @@ Object.assign(india_compliance, {
         if (UNBODY_REGEX.test(gstin)) return "UIN Holders";
         if (OVERSEAS_REGEX.test(gstin)) return "Overseas";
     },
+
+    set_hsn_code_query(field) {
+        if (!field || !gst_settings.validate_hsn_code) return;
+        field.get_query = function () {
+            const wildcard = '_'.repeat(gst_settings.min_hsn_digits) + '%';
```
ankush, line 135: Nice. Didn't know this was used for limiting characters.

PR #1042, opened 2023-09-11
feat: Click to Filter in Purchase Reconciliation Tool

vishakhdesai:


sagarvora (changes requested):
let's remove `filter_button_group`, `filter_button` & `filter_x_button` from `this`. They aren't used outside the `setup_filter_button` function.

sagarvora:
rest looks ok.

sagarvora (approved):

PR #1046, opened 2023-09-13
fix: purchase reconciliation fixes

ljain112:


india_compliance/gst_india/overrides/sales_invoice.py
```diff
@@ -214,6 +214,7 @@ def update_dashboard_with_gst_logs(doctype, data, *log_doctypes):
         {
             "e-Waybill Log": "reference_name",
             "Integration Request": "reference_docname",
+            "GST Inward Supply": "link_name",
```
vorasmit, line 217: Is this required in the Sales Invoice?

ljain112: This is a utility function,this is called in purchase invoice also.just updating non standard field.

vorasmit:
@Mergifyio backport version-14-hotfix

PR #1059, opened 2023-09-18
fix: common place of supply options for all doctypes

ljain112:
Internal Reference - https://support.frappe.io/app/hd-ticket/1333


vorasmit:
Govt calls it `Other Country`.
Is rename good to have?

vorasmit:

PR #1080, opened 2023-09-26
chore: Hook for default company account fields

deepeshgarg007:
Required for: https://github.com/frappe/erpnext/pull/37238


vorasmit:
Company on_update hooks already handle this.
https://github.com/resilient-tech/india-compliance/blob/develop/india_compliance/hooks.py#L76
https://github.com/resilient-tech/india-compliance/blob/develop/india_compliance/gst_india/overrides/company.py#L158

This may be required conditionally where the country is India and this hook may not do it correctly for foreign companies.
This is best handled from regional or relevant apps in my opinion.

PR #1113, opened 2023-10-09
fix(ui): show sandbox mode indicator in transaction doctypes

### LINKED ISSUE #1055: Sandbox Mode status for frontend in Transactions (like: Experimental)
### Suggested Implementation

Sidebar in Transactions

image_2023-07-28_15-49-29

DaizyModi:
image


india_compliance/public/js/transaction.js
```diff
@@ -22,6 +23,31 @@ for (const doctype of ["Sales Invoice", "Delivery Note"]) {
     ignore_port_code_validation(doctype);
 }
 
+function show_sandbox_mode_indicator(doctype) {
```
vorasmit: Keeping this in e-Waybill Actions would restrict this to relevant doctypes.

PR #1135, opened 2023-10-11
fix: set gst_category in party and address doctype

### LINKED ISSUE #1063: Set correct GST category if not set
https://github.com/resilient-tech/india-compliance/blob/9224fd882a6171cf5390f10e314028a4c3049526/india_compliance/gst_india/overrides/address.py#L21-L31

This validation breaks integrations, can't we just set the right category if it's not set here? I get throwing error on incorrect value but when it's unset there's only one sane way to move forward, so we might as well just do it in code.

ljain112:


india_compliance/gst_india/client_scripts/address.js
```diff
@@ -18,7 +18,13 @@ frappe.ui.form.on(DOCTYPE, {
         // Automatically set GST Category as Overseas if country is not India
         if (frm.doc.country != "India") {
             frm.set_value("gst_category", "Overseas");
+            return;
         }
+
+        frm.set_value(
```
vorasmit: trigger gstin

india_compliance/gst_india/overrides/party.py
```diff
@@ -5,19 +5,69 @@
...
+        _("GST Category has been updated to {0} based on GSTIN.").format(
+            frappe.bold(gst_category)
+        ),
+        indicator="green",
+        alert=True,
+    )
+
+
+def fetch_or_guess_gst_category(doc):
+    gst_category = guess_gst_category(doc.gstin, doc.get("country"))
```
vorasmit: In else cond

india_compliance/gst_india/overrides/party.py
```diff
@@ -5,19 +5,69 @@
...
+
+    if doc.gstin and not doc.gst_category and is_autofill_party_info_enabled():
```
vorasmit: Handle unregistered

PR #1165, opened 2023-10-16
fix: reset hsn with item code

GursheenK:
### Reverted in favor of https://github.com/resilient-tech/india-compliance/pull/1174

**Bug**
For all the DocTypes that use the Items child table, **HSN / SAC** field is not reset when the Item code changes.

<br>

https://github.com/resilient-tech/india-compliance/assets/40693548/de64808e-ba99-4521-9143-3a5b316e08bb

<br>

**Fix**

- Disabled the `fetch_if_empty` flag for the `gst_hsn_code` field in the list of custom fields.
- Added a patch to update the flag for already existing custom field.


sagarvora:
This fix is only required because of an incorrect fix in Frappe:
https://github.com/frappe/frappe/pull/21643

PR #1393, opened 2023-12-12
chore: optional show_msg for asset rate change

anandbaburajan:
Showing the message only when needed

Required for https://github.com/frappe/erpnext/pull/38697.


vorasmit:
Backported #1392

PR #1412, opened 2023-12-15
fix: modify params as per erpnext

ljain112:
Depends On: frappe/erpnext#38744


vorasmit:

PR #1421, opened 2023-12-18
fix: remove extra `validate` parameter

rtdany10:
image

Unable to fetch outstanding invoices in v14


vorasmit:

PR #1438, opened 2023-12-19
fix: update taxable in patch, continue without gst_accounts

vorasmit:


india_compliance/patches/post_install/improve_item_tax_template.py
ljain112, line 1: :133 template_name

ljain112: :53 constant not used

PR #1509, opened 2023-12-30
fix: don't execute patch if column is absent or missing gst accounts

vorasmit:


india_compliance/patches/post_install/improve_item_tax_template.py
```diff
@@ -52,6 +52,9 @@
 
 
 def execute():
+    if not frappe.db.has_column("Item", "is_nil_exempt"):
+        return
```
ljain112, line 56: update_gst_details_for_transaction can run multiple times.

vorasmit: The column is still present in the database even after we delete the custom field. This is to ensure the patch doesn't run for a completely fresh site.

PR #1510, opened 2023-12-30
refactor: company tax setup in setup wizard

vorasmit:
## Fixes

- [x] Consistent allowed tax rates as string
- [x] Update Trigger event for company name as updated in ERPNext
- [x] Remove unnecessary stage for setting up tax template


india_compliance/setup_wizard.py
ljain112, line 1: :104 not required already checking in main function

PR #1521, opened 2023-12-31
fix: asset WDV depreciation calc according to IT act [dev]

anandbaburajan:
First year depreciation for WDV method as per Income Tax act wasn't being calculated properly. Now the depreciation calculations in ERPNext match these [calculations](https://docs.google.com/spreadsheets/d/1zD33bA5kyNmrajNpMxVmMao5eio0a09B/edit?usp=sharing&ouid=105035438770380117661&rtpof=true&sd=true) provided by a customer.

Needs to be merged with: https://github.com/frappe/erpnext/pull/39052.

### DIFF
--- a/india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
+++ b/india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
@@ -1,29 +1,132 @@
 import frappe
 from frappe import _
-from frappe.utils import date_diff
+from frappe.utils import (
+    add_days,
+    add_months,
+    cint,
+    date_diff,
+    flt,
+    get_last_day,
+    getdate,
+    is_last_day_of_the_month,
+)
+from erpnext.accounts.utils import get_fiscal_year
+from erpnext.assets.doctype.asset_depreciation_schedule.asset_depreciation_schedule import (
+    get_asset_depr_schedule_doc,
+    get_default_wdv_or_dd_depr_amount,
+)
 
 
-def get_updated_rate_of_depreciation_for_wdv_and_dd(
-    asset, depreciable_value, fb_row, show_msg=True
+def get_wdv_or_dd_depr_amount(
+    asset,
+    fb_row,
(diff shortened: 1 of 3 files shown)


nabinhait (approved):

nabinhait (approved):

vorasmit:
## Replicating example
Looks like the dep for 2nd year onwards is incorrect.

Asset Config (with Income Tax Book). Eg with 15% WDV for Income Tax
image


### Depreciation with Annual Frequency
image

### Depreciation with Monthly Frequency
image

### Issues
- As per monthly depreciation, dep for the second year comes to 20459.82-7500 = 12959.82. This doesn't equal 13875 for the annum.
- Method of depreciation should be applied for the financial year. However, monthly depreciation can be simpler and should be distributed equally based on the number of months or days. Currently, it's wdv every month, which is not required.

## Proposal =>
Dep calculation could be broken down for simplicity IMO,

1) Calculate Full Years Dep
2) Regional Override vs ERPNext Functionality to calculate proportion for the year
3) Distribute the proportionate over multiple periods within the year (ideally equally based on the of days)

So irrespective of depreciation periodicity and method, depreciation could always be calculated for the year, and then distributed in equal proportion for monthly/quarterly MIS or reporting.

india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
```diff
@@ -1,29 +1,132 @@
...
+        return get_default_wdv_or_dd_depr_amount(
+            asset,
+            fb_row,
+            depreciable_value,
+            schedule_idx,
+            prev_depreciation_amount,
+            has_wdv_or_dd_non_yearly_pro_rata,
+        )
+
+    if not fb_row.daily_prorata_based:
```
vorasmit, line 43: Customers example for pro rata on a daily basis is a personal preference and not a requirement as per income tax act.

PR #1603, opened 2024-01-15
WDV Depreciation as per income tax act

nabinhait:
Depends on https://github.com/frappe/erpnext/pull/39385


vorasmit:
## Daily Pro-Rata for first month looks incorrect

When available from the use date is in the middle of the month, daily pro-rata is still considered from the start of the month.

Available from use date:
image

Depreciation for the first year:
image

vorasmit (dismissed):

vorasmit (approved):

PR #1662, opened 2024-02-05
fix: reduce complexity for counting GST `tax rate` and `amount` when generating `e_invoice`

Abdeali099:
## WORK

1. Added GST  `tax rate` and `amount` directly  into `get_all_item_details()`
2.  Removed `for..loop` code from `update_item_tax_details()` which counting `tax rate` and `amount`

## CHANGED CODE

https://github.com/resilient-tech/india-compliance/blob/76434a30a089e7447c9d5a1d2d2b4e283011eda2/india_compliance/gst_india/utils/transaction_data.py#L273-L294


vorasmit:
This shall be done with a complete revamp to usage of taxes at all places.

PR #1674, opened 2024-02-13
fix: remove gstin fields for non-indian countries in setup wizard

divyam-mistry:
- India specific fields

**For non-Indian countries:**

image

**For India:**

image

- Validate GSTIN

**Invalid GSTIN**

image

### DIFF
--- a/india_compliance/public/js/setup_wizard.js
+++ b/india_compliance/public/js/setup_wizard.js
@@ -10,9 +10,7 @@ frappe.setup.on("before_load", function () {
     first_slide.onload = function (slide) {
         _onload.call(this, slide);
 
-        const country_input = frappe.wizard
-            ? frappe.wizard.slide_dict[0].get_input("country")
-            : null;
+        const country_input = frappe.wizard?.slide_dict[0].get_input("country");
         if (country_input) {
             country_input.on("change", event => {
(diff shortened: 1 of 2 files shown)


india_compliance/setup_wizard.py
```diff
@@ -55,6 +59,11 @@ def setup_company_taxes(params):
     if not (params.company_name and frappe.db.exists("Company", params.company_name)):
         return
 
+    try:
+        validate_gstin(params.company_gstin)
+    except Exception:
```
vorasmit: can we instead use frappe.ValidationError?

india_compliance/public/js/setup_wizard.js
```diff
@@ -65,6 +104,15 @@ function update_erpnext_slides_settings() {
 async function autofill_company_info(slide) {
     const gstin = slide.get_input("company_gstin").val();
     const gstin_field = slide.get_field("company_gstin");
+
+    set_gstin_description(gstin_field);
```
vorasmit: We don't want this in the setup wizard.
It should be enough to have a red-bordered input box, with small text in red saying invalid GSTIN (only if it's invalid)

divyam-mistry: Used `field_obj.$input.css("border", "1px solid red");`

PR #1791, opened 2024-02-28
ci: allow workflow call for other projects to use

Sanket322:
This helps in reusing workflows


.github/workflows/server-tests.yml
```diff
@@ -1,6 +1,13 @@
 name: Server Tests
 
 on:
+    workflow_call:
+      inputs:
+        base_ref:
+          type : string
+      secrets:
+        codecov_token:
+          required: true
```
vorasmit: Let's make codecov optional.
If this is available, only then we upload coverage data.

PR #1811, opened 2024-03-01
perf: optimise patch `update_gst_details_for_invoices_with_same_item_multiple_times`

sagarvora:
- get only GST accounts with values, i.e. not None

- attempted implementation of https://stackoverflow.com/questions/6135376/mysql-select-where-field-in-subquery-extremely-slow-why (but doesn't seem to improve anything on SMC test)

pypika doc: https://pypika.readthedocs.io/en/latest/2_tutorial.html#with-clause


vorasmit:
StackOverflow article is about MySQL.
Tried it myself again and could not identify any significant performance improvement.

This is used at times for readability, which is already available easily with pypika.

This could be useful if more operations were required on aliased queries, like grouping etc.

PR #1832, opened 2024-03-11
fix(minor): Show a message if removing invalid GSTIN

rmehta:
Was testing some thing with a sandbox GSTIN - got very annoyed that it kept deleting it without any message.
This will at least let the user know their GSTIN is invalid.


rmehta:
Thanks @vorasmit @sagarvora :-)

PR #1833, opened 2024-03-11
fix: ensure links exist before setting default values for GST fields in Address

Ninad1306:


india_compliance/gst_india/client_scripts/address.js
```diff
@@ -25,7 +25,7 @@ frappe.ui.form.on(DOCTYPE, {
         india_compliance.set_state_options(frm);
 
         // set default values for GST fields
-        if (!frm.is_new() || !frm.doc.links || frm.doc.gstin) return;
+        if (!frm.is_new() || !frm.doc.links.length || frm.doc.gstin) return;
```
sagarvora:
```suggestion
        if (!frm.is_new() || !frm.doc.links || !frm.doc.links.length || frm.doc.gstin) return;
```

PR #1883, opened 2024-03-20
fix: set gstin and gstCategory

Sanket322:
set gstin and gst_category of customer/supplier when saving the address

### DIFF
--- a/india_compliance/gst_india/overrides/address.py
+++ b/india_compliance/gst_india/overrides/address.py
@@ -9,23 +9,45 @@ from india_compliance.gst_india.utils import (
     validate_pincode,
 )
 
-def set_gstin_and_gstCategory(doc, method=None):
-    links = doc.links
 
-    for link in links:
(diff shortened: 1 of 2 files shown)


india_compliance/gst_india/overrides/address.py
```diff
@@ -9,6 +9,24 @@
...
+        if doctype not in ['Customer','Supplier']:
+            continue
+
+        gstin = doc.gstin
+        if gstin:
+            frappe.db.set_value(doctype,docname,"gstin",gstin)
+
+        gst_category = doc.gst_category
+        if gst_category:
+            frappe.db.set_value(doctype, docname, "gst_category", gst_category)
```
vorasmit: only set if gstin exists and gst_category in party (customer/supplier) is unregistered. 
Alert => GSTIN updated in Party (optional)

vorasmit (changes requested):

vorasmit (approved):

PR #1986, opened 2024-04-06
fix(minor): ReferenceError: taxes_table is not defined in Journal Entry

vorasmit:


vorasmit:
https://github.com/Mergifyio backport develop

Issue #2072, opened 2024-04-26
Inconsistency with GST Transporter ID


ljain112:
Frappe Support Issue: https://support.frappe.io/app/hd-ticket/13855
![image](https://github.com/resilient-tech/india-compliance/assets/108322669/809e1057-5b80-4428-b65b-392726a3c126)

Eg Transporter ID: 29AAFCA7488L1Z0
This is a valid transporter ID but Invalid GSTIN and check digit validation is failing.

### Proposed Solution 
- Avoid check digit validation for all transporter ID and only validate GSTIN regex
- Use directly transporter ID API for all the gstin with transporter or Remove transporter validation at all.

PR #2080, opened 2024-04-26
revert: GST Transporter ID API validation (refactor and fixes required)

vishakhdesai:
Temporary fix to the issue: #2072


vorasmit:
Instead we can refactor this.

PR #2109, opened 2024-05-10
fix: include a support email for API-related errors.

Sanket322:
When an error related to the India Compliance API occurs, you have the option to email API Support with the necessary information.


vorasmit:
Instead better way to this is through auto-retry.

PR #2117, opened 2024-05-14
fix: support custom print format for e-Waybill

vorasmit:


india_compliance/gst_india/utils/e_waybill.py
```diff
@@ -717,10 +717,12 @@ def mark_e_waybill_as_cancelled(doctype, docname, values):
 
 
 def attach_e_waybill_pdf(doc, log=None):
+    default_format = frappe.get_meta("e-Waybill Log").default_print_format
```
sagarvora, line 720: this isn't required at all. if you don't specify any print format, this is what gets chosen.

india_compliance/gst_india/setup/property_setters.py
```diff
@@ -111,6 +105,21 @@ def get_property_setters():
     ]
 
 
+def get_default_property_setters():
```
sagarvora, line 108: why is this a function. I don't see any kind of computation happening here. recommending a constant like `TRANSPORTER_NAME_PROPERTIES` instead. suggested constant name: `DEFAULT_PROPERTIES`.

it should then be optionally sent with the return value of `get_property_setters` based on a flag called something like `include_defaults` (flag should default to False).

Issue #2093, opened 2024-05-02
Error in gstin info


ljain112:
Frappe Support Issue: https://support.frappe.io/app/hd-ticket/14349

```
App Versions

{
    "abk_customization": "0.0.1",
    "erpnext": "15.21.2",
    "frappe": "15.25.0",
    "hrms": "15.20.1",
    "india_compliance": "15.7.3",
    "insights": "2.2.2",
    "sales_application_plugin": "0.0.1"
}

Route

List/Customer/List

Traceback

Traceback (most recent call last):
  File "apps/frappe/frappe/app.py", line 110, in application
    response = frappe.api.handle(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/api/__init__.py", line 49, in handle
    data = endpoint(**arguments)
           ^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/api/v1.py", line 36, in handle_rpc_call
    return frappe.handler.handle()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/handler.py", line 49, in handle
    data = execute_cmd(cmd)
           ^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/handler.py", line 85, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/__init__.py", line 1761, in call
    return fn(*args, **newargs)
           ^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/utils/typing_validations.py", line 31, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/desk/form/save.py", line 39, in savedocs
    doc.save()
  File "apps/frappe/frappe/model/document.py", line 337, in save
    return self._save(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/model/document.py", line 373, in _save
    self.run_before_save_methods()
  File "apps/frappe/frappe/model/document.py", line 1091, in run_before_save_methods
    self.run_method("validate")
  File "apps/frappe/frappe/model/document.py", line 962, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
[...]

PR #2123, opened 2024-05-16
fix: convert into titlecase only if value exists

nabinhait:
Screenshot 2024-05-16 at 2 49 47 PM


vorasmit:
- City and state is mandatory and should always exist in the autofill API.
- There was one issue that ignored city for duplicates #2093. Was already fixed.

Updates should help.

PR #2144, opened 2024-05-20
fix: use base_tax_amount_after_discount_amount instead of tax_amount

### LINKED ISSUE #2143: Should we use tax_amount_after_discount_amount instead of tax amount for calculations ?
https://github.com/resilient-tech/india-compliance/blob/9c3341c758ec1c783b695004a4de3345e17fa1d5/india_compliance/gst_india/overrides/transaction.py#L1044

ljain112:
Closes: #2143 

Tax difference was calculated on document currency but taxes were set on the basis of company currency.
image

### DIFF
--- a/india_compliance/gst_india/utils/jinja.py
+++ b/india_compliance/gst_india/utils/jinja.py
@@ -206,11 +206,11 @@ class GSTBreakup:
                     tax_type,
                     {
                         "tax_rate": flt(item.get(f"{_tax_type}_rate", 0)),
-                        "tax_amount": 0,
+                        "base_tax_amount_after_discount_amount": 0,
                     },
                 )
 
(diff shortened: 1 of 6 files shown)


india_compliance/gst_india/utils/jinja.py
```diff
@@ -206,11 +206,11 @@ def get(self):
                     tax_type,
                     {
                         "tax_rate": flt(item.get(f"{_tax_type}_rate", 0)),
-                        "tax_amount": 0,
+                        "base_tax_amount_after_discount_amount": 0,
```
vorasmit: Is this required? This looks like internal variable.
Moreover, this used GST Tax details from items.

vorasmit (approved):

PR #2151, opened 2024-05-21
fix: add description and hide progress bar in purchase reconciliation tool

Sanket322:
Add description when `Purchase Period` or `Inward Supply Period` is selected as `Last Fiscal year` in purchase reconciliation tool.
Remove progress bar if it is there after completion of function


india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.js
```diff
@@ -143,9 +149,14 @@ frappe.ui.form.on("Purchase Reconciliation Tool", {
 
     purchase_period(frm) {
         fetch_date_range(frm, "purchase");
+        let description = frm.doc.purchase_period == "Last Fiscal Year" ? fiscal_start +' to '+ fiscal_end : ""
+        frm.get_field("purchase_period").set_description(description);
```
vorasmit: This should be for all periods. Date range is already fetched. We can use directly that to show in description.

PR #2159, opened 2024-05-23
fix: resolved discrepancy in function arguments for depreciation calcu…

khushi8112:
This pull request addresses an issue where a function override caused by the regional compliance module results in a TypeError during asset creation.
The "get_wdv_or_dd_depr_amount" function, modified in erpnext, is overridden by india_compliance, resulting in a mismatch in the number of arguments.

**Fix**
Modified the india_compliance function to match the signature expected by erpnext, ensuring consistency in the number and type of arguments.

### DIFF
--- a/india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
+++ b/india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
@@ -42,7 +42,7 @@ def get_wdv_or_dd_depr_amount(
             prev_depreciation_amount,
             has_wdv_or_dd_non_yearly_pro_rata,
             asset_depr_schedule,
-            prev_per_day_depr=0,
+            prev_per_day_depr=prev_per_day_depr,
         )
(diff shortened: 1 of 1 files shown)


vorasmit:
Reference ERPNext PR:
https://github.com/frappe/erpnext/pull/41235

india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
```diff
@@ -41,6 +42,7 @@ def get_wdv_or_dd_depr_amount(
             prev_depreciation_amount,
             has_wdv_or_dd_non_yearly_pro_rata,
             asset_depr_schedule,
+            prev_per_day_depr=0,
```
vorasmit:
```suggestion
            prev_per_day_depr=prev_per_day_depr,
```

vorasmit (approved):

PR #2194, opened 2024-05-28
fix: Dynamic Link for Purchase Doc in Purchase Reco

ljain112:
image


india_compliance/public/js/purchase_reconciliation_tool/data_table_manager.js
```diff
@@ -78,6 +78,11 @@ india_compliance.DataTableManager = class DataTableManager {
                 value = column._value(value, column, data);
             }
 
+            // format Dynamic Link as a Link
+            if (column.fieldtype === "Dynamic Link") {
+                column.docfield = { fieldtype: "Link", options: data[column.options] };
+            }
+
```
vorasmit:
```suggestion
            return frappe.format(value, row, column, data)

```

PR #2379, opened 2024-07-10
fix: section 9(5) of 3.1.1 of gstr_3b

priyanshshah2442:


vorasmit:
- [x] Are 3.1 (a) reverse charge  should be without taxes 
- [x] Reduce taxable value to extent of reverse charge through ecom
- [x] common query and better function names
- [x] test case

PR #2393, opened 2024-07-11
fix: set party type from pan no

### LINKED ISSUE #2052: Can customer type/supplier type can be set based on GST Search API ?
https://support.frappe.io/app/hd-ticket/13606

Sanket322:
Closes : #2052

### DIFF
--- a/india_compliance/gst_india/client_scripts/party.js
+++ b/india_compliance/gst_india/client_scripts/party.js
@@ -70,9 +70,7 @@ function validate_gstin(doctype) {
             if (PAN_REGEX.test(pan)) {
                 frm.doc.pan = pan;
                 frm.refresh_field("pan");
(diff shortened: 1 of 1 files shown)


vorasmit:
We can patch party type proprietorship in ERPNext. Use Individual instead.
@ljain112 Thoughts?

india_compliance/gst_india/client_scripts/party.js
```diff
@@ -70,6 +70,9 @@ function validate_gstin(doctype) {
             if (PAN_REGEX.test(pan)) {
                 frm.doc.pan = pan;
                 frm.refresh_field("pan");
+                if(["Customer", "Supplier"].includes(doctype)){
```
vorasmit: Can be common condition for `set_party_type`. You can access doctype as `frm.doc.doctype`

PR #2555, opened 2024-08-23
fix: New GST Category `Input Service Distributor` Introduced

Ninad1306:
This new category is useful for automatic classification of ITC categories for GSTR-3b


vorasmit:
- [x] Warning for users already using ISD classification.

PR #2589, opened 2024-09-02
fix: changed `doctype` column in audit trail report from Link to Data

ljain112:
Only the System Manager has permission to read doctype "Doctype".
Also, frappe doesn't allow giving permission for Doctype.

https://support.frappe.io/app/hd-ticket/21253


vorasmit:
Rather this could be autocomplete with the list of audit trail doctypes. Keeping it as data field will affect the UX.

ljain112:
How this will affect UX?
 Dynamic Links will be working.

PR #2594, opened 2024-09-04
fix: Extended E-Way Bill Support for Stock Entry

Ninad1306:


india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -69,11 +69,18 @@ def after_mapping_subcontracting_order(doc, method, source_doc):
 
 
 def after_mapping_stock_entry(doc, method, source_doc):
-    if source_doc.doctype == "Subcontracting Order":
+    if not source_doc.doctype == "Subcontracting Order":
+        doc.taxes_and_charges = ""
+        doc.taxes = []
+
+    if not doc.purpose == "Material Transfer" or not doc.is_return:
```
Ninad1306: This will not work standalone, as some fields in erpnext has to be updated in postprocess to correctly evaluate the conditions.

india_compliance/gst_india/constants/custom_fields.py
```diff
@@ -181,14 +181,14 @@
             "label": "Taxes",
             "fieldtype": "Section Break",
             "insert_after": "get_stock_and_rate",
-            "depends_on": "eval:doc.subcontracting_order",
+            "depends_on": "eval:((doc.purpose == 'Send to Subcontractor' && doc.subcontracting_order) || (in_list(['Material Transfer', 'Material Issue'], doc.purpose)))",
```
vorasmit: india_compliance utility for readability

india_compliance/gst_india/utils/e_waybill.py
```diff
@@ -1264,7 +1263,10 @@ def validate_applicability(self):
         self.validate_same_gstin()
 
     def validate_same_gstin(self):
-        if self.doc.doctype == "Delivery Note":
+        if self.doc.doctype == "Delivery Note" or (
+            self.doc.get("purpose") in ["Material Transfer", "Material Issue"]
+            and not self.doc.is_return
```
vorasmit: possibility to use subcontracting order field

india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -69,11 +69,18 @@ def after_mapping_subcontracting_order(doc, method, source_doc):
 
 
 def after_mapping_stock_entry(doc, method, source_doc):
-    if source_doc.doctype == "Subcontracting Order":
+    if not source_doc.doctype == "Subcontracting Order":
```
vorasmit: !=

india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -165,7 +172,7 @@ def onload(doc, method=None):
 
 
 def validate(doc, method=None):
-    if ignore_gst_validation_for_subcontracting(doc):
+    if ignore_based_on_purpose(doc):
```
vorasmit: cannot have e-Waybill

india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -390,3 +402,13 @@ def remove_duplicates(doc):
         doc.doc_references = []
         for row in references:
             doc.append("doc_references", dict(link_doctype=row[0], link_name=row[1]))
+
+
+def ignore_based_on_purpose(doc):
+    if doc.doctype == "Stock Entry" and (
```
vorasmit: simplify

india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -69,11 +69,18 @@ def after_mapping_subcontracting_order(doc, method, source_doc):
...
+
+    if not doc.purpose == "Material Transfer" or not doc.is_return:
         return
 
-    doc.taxes_and_charges = ""
-    doc.taxes = []
+    doc.bill_to_address = source_doc.billing_address
+    doc.bill_from_address = source_doc.supplier_address
+    doc.bill_to_gstin = source_doc.company_gstin
+    doc.bill_from_gstin = source_doc.supplier_gstin
```
vorasmit, line 89: test case for same gstin transfer

india_compliance/gst_india/utils/e_waybill.py
Ninad1306: Subcontracting order is not a mandatory field in Stock Entry and also is not mandatory for e-Waybill Generation.
So using it may not be a differentiating factor.

india_compliance/gst_india/overrides/subcontracting_transaction.py
Ninad1306: Reference PR in ERPNext: https://github.com/frappe/erpnext/pull/43061

india_compliance/gst_india/utils/e_waybill.py
```diff
@@ -1267,6 +1267,9 @@ def validate_same_gstin(self):
         if self.doc.doctype == "Delivery Note":
             return
 
+        if is_outward_material_transfer(self.doc):
```
vorasmit: Are we ensuring if GSTIN is same?

india_compliance/gst_india/constants/custom_fields.py
```diff
@@ -181,14 +181,14 @@
             "label": "Taxes",
             "fieldtype": "Section Break",
             "insert_after": "get_stock_and_rate",
-            "depends_on": "eval:doc.subcontracting_order",
+            "depends_on": "eval: india_compliance.is_e_waybill_generatable_for_subcontracting(doc)",
```
vorasmit, line 185: This condition should be consistent with backend condition to calculate taxes.

india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -314,8 +329,11 @@ def validate(self, doc, is_sales_transaction=False):
         self.validate_for_charge_type()
 
     def validate_for_same_party_gstin(self):
```
vorasmit, line 331: after fixing `is_e_waybill_generatable_for_subcontracting`, this function may become redundant.

india_compliance/public/js/transaction.js
```diff
@@ -136,7 +142,17 @@ async function update_gst_details(frm, event) {
...
     } else if (frm.doc.doctype === "Stock Entry") {
-        fieldnames_to_set.push("bill_from_gstin", "bill_to_gstin", "bill_to_address");
+        fieldnames_to_set.push(
+            "bill_from_gstin",
+            "bill_to_gstin",
+            "bill_from_address",
+            "bill_to_address"
+        );
+
+        party_details["is_outward_material_transfer_or_issue"] =
```
vorasmit: consistent flag name `is_outward_stock_entry` `is_inward_stock_entry`

vorasmit: Can we assume its inward, if it's not outward?

india_compliance/public/js/transaction.js
```diff
@@ -79,7 +79,13 @@ async function update_gst_details(frm, event) {
     const party_type = india_compliance.get_party_type(frm.doc.doctype).toLowerCase();
     const party_fieldname = frm.doc.doctype === "Quotation" ? "party_name" : party_type;
     const party = frm.doc[party_fieldname];
-    if (!party) return;
+
+    const is_outward_material_transfer_or_issue =
```
vorasmit: more readable check I feel is:
same_gstin_stock_entry

Party or same_gstin_stock_entry should be there.

india_compliance/public/js/transaction.js
Ninad1306: It would be too general as this function is called for all three doctypes. So if its not outward stock enrty then it may be a Subcontracting receipt or Subcontracting order or an inward stock entry.

PR #2622, opened 2024-09-17
fix: make `e-Invoice log` extensible

Sanket322:


india_compliance/patches/v14/make_e_invoice_log_extensible.py
```diff
@@ -0,0 +1,9 @@
+import frappe
```
ljain112: move file to v15 folder

ljain112 (approved):

PR #2647, opened 2024-09-30
fix: pos based on account settings

Sanket322:


india_compliance/gst_india/overrides/transaction.py
```diff
@@ -1493,13 +1498,24 @@ def validate_transaction(doc, method=None):
 
     validate_overseas_gst_category(doc)
 
+    address = frappe.db.get_single_value(
```
ljain112: variable_name

india_compliance/gst_india/utils/__init__.py
```diff
@@ -390,6 +390,10 @@ def get_place_of_supply(party_details, doctype):
     # fallback to company GSTIN for sales or supplier GSTIN for purchases
     # (in retail scenarios, customer / company GSTIN may not be set)
 
+    determine_address_tax_category_from = frappe.db.get_single_value(
```
vorasmit: cached value

vorasmit (approved):

PR #2652, opened 2024-10-07
fix: only show warning for subcontracting receipt

Ninad1306:
For subcontracting, users could have different workflows:

eg: Material sent to subcontractor directly from the supplier's warehouse.
Because of these reasons, it becomes difficult to add references.

With a warning, this document can be skipped in ITC-04.


vorasmit:
Can we add test cases for them?

PR #2656, opened 2024-10-09
fix: add comment in gst return log for backdated transactions

Sanket322:


india_compliance/gst_india/overrides/transaction.py
```diff
@@ -660,17 +663,55 @@ def get_source_state_code(doc):
...
+    if not gstr_1_filed_upto or getdate(doc.posting_date) > gstr_1_filed_upto:
+        return
+
+    doc_name = get_gst_return_log_name(doc)
+    if not doc_name:
+        return
+
+    gst_return_log = frappe.get_doc("GST Return Log", doc_name)
+
+    url = f"{frappe.utils.get_url()}/app/{doc.doctype.replace(' ', '-').lower()}/{doc.name}"
```
ljain112: get_link_to_form

india_compliance/gst_india/overrides/transaction.py
```diff
@@ -660,17 +663,55 @@ def get_source_state_code(doc):
...
+    gst_return_log.add_comment(
+        "Comment",
+        f"{doc.doctype} : [{doc.name}]({url}) has been {action} by {frappe.session.user}",
+    )
+
+
+def get_gst_return_log_name(doc):
+    posting_date = getdate(doc.posting_date)
```
ljain112: default last log

india_compliance/gst_india/overrides/transaction.py
```diff
@@ -660,17 +663,55 @@ def get_source_state_code(doc):
...
+    year = posting_date.year
+    month = f"{posting_date.month:02d}"
+
+    doc_name = f"GSTR1-{month}{year}-{doc.company_gstin}"
```
ljain112: use filter instead of naming

india_compliance/gst_india/overrides/transaction.py
vorasmit: Objective was to make sense of differences after return is filed. Hence ideally comment should form part of the period where difference will be visible in reconciliation.

vorasmit (approved):

PR #2666, opened 2024-10-15
fix: connection for e-Invoice log in purchase invoice

Sanket322:


india_compliance/gst_india/overrides/purchase_invoice.py
```diff
@@ -166,6 +166,7 @@ def get_dashboard_data(data):
         "e-Waybill Log",
         "Integration Request",
         "GST Inward Supply",
+        "e-Invoice Log",
```
vorasmit: Can we have sequence same as Sales Invoice?

PR #2685, opened 2024-10-25
fix: HSN Code disappears in `Edit Full Form`

Ninad1306:
Issue Reference: https://support.frappe.io/helpdesk/tickets/24124


vorasmit:
Should be fixed in Frappe.

PR #2691, opened 2024-11-05
fix: improve ux for Journal Entry Creation

Sanket322:


india_compliance/gst_india/doctype/gstr_1_beta/gstr_1_beta.js
```diff
@@ -715,10 +715,82 @@ class GSTR1 {
...
+            method: "india_compliance.gst_india.doctype.gstr_1_beta.gstr_1_beta.get_journal_entries",
+            args: { month_or_quarter, year , company },
+        })
+
+        if(!data.message) return;
+
+        const dialog = this.create_journal_entry_dialog(data.message)
+        dialog.show();
+
+        const journal_entry_html = dialog.get_field("journal_entry_accounts")
```
vorasmit: can be directly updated.

PR #2772, opened 2024-11-26
fix: do not check gst category for setting Place of supply

ljain112:


ljain112:
@vorasmit backport version-14

PR #3049, opened 2025-02-11
fix: refresh GSTIN status only when creating transactions

vishakhdesai:
- Validation in **GST Settings**: Throw error if **GSTIN Status Refresh Interval** is less than 15
- Patch to make **GSTIN Status Refresh Interval** 15 if it is less than 15
- Do not call GSTIN search API on refresh of submitted or cancelled transactions
- Do not call GSTIN search API on refresh of Party doctype


india_compliance/patches.txt
```diff
@@ -69,3 +69,4 @@ india_compliance.patches.v15.set_default_for_new_gst_category_notification
 india_compliance.patches.v15.make_e_invoice_log_extensible
 india_compliance.patches.v15.migrate_boe_taxes_to_ic_taxes
 india_compliance.patches.v15.set_download_document_for_2a_2b
+india_compliance.patches.v15.update_gstin_status_refresh_interval
```
vorasmit: we can backport this to v14 too. Keep it in v14 folder.

india_compliance/gst_india/doctype/gst_settings/gst_settings.py
```diff
@@ -314,6 +315,17 @@ def validate_e_invoice_applicable_companies(self):
...
+    def validate_gstin_status_refresh_interval(self):
+        if not (
+            self.enable_api
+            and self.validate_gstin_status
+            and self.gstin_status_refresh_interval is not None
+        ):
+            return
+
+        if self.gstin_status_refresh_interval < 15:
+            frappe.throw(_("GSTIN Status Refresh Interval cannot be less than 15 days"))
```
vorasmit: Let's just set it to 15 without raising error.
Add description to this field: 15 days or above.

PR #3153, opened 2025-03-06
test: Incorrect Purchase Reconciliation Tool Test Case and Data

Ninad1306:


india_compliance/gst_india/doctype/purchase_reconciliation_tool/test_purchase_reconciliation_tool.py
```diff
@@ -95,7 +95,7 @@ def test_purchase_reconciliation_tool(self):
         )
 
         purchase_reconciliation_tool.save(ignore_permissions=True)
```
vorasmit: save may not be necessary

PR #3168, opened 2025-03-08
fix: ignore item wise tax validation when dont_recompute_tax is enabled

venkat102:
**Issue:**
Unable to sync orders from e-commerce sites.
ref: [33079](https://support.frappe.io/helpdesk/tickets/33079)

**Solution:**
As the tax details fetched from the e-commerce sites won't be recalculated, the validation can be ignored if `dont_recompute_tax` is enabled.

image

Backport needed for v15 & v14

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -147,6 +147,9 @@ def validate_item_wise_tax_detail(doc):
         if not row.gst_tax_type:
             continue
 
+        if row.dont_recompute_tax:
+            continue
+
         if row.charge_type != "Actual":
             continue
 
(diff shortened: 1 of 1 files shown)


vorasmit:
Hi, This would need a fix in Shopify App. This is because if not handled correctly, this would lead to incorrect GST Details in Item => Incorrect e-Invoice and/or Incorrect GSTR-1

Could you check for the reason why `tax_difference` is coming [here](https://github.com/resilient-tech/india-compliance/pull/3168/files#diff-daf9ff43d80eee4a7b5733e3aee0137865b04f4c6296d765a0e7a2537f165aedR184)?

vorasmit:
This function is introduced to validate taxes where `don't recompute tax` is checked. ensures GST values set are calculated correctly. Returning early defeats the purpose of the function.

PR #3184, opened 2025-03-11
fix: Incorrect Invoice Totals for GSTR-1 Beta

Ninad1306:


india_compliance/gst_india/report/gst_sales_register_beta/gst_sales_register_beta.py
```diff
@@ -302,6 +302,13 @@ def get_columns(filters):
                 "fieldtype": "Currency",
                 "options": company_currency,
             },
+            {
+                "label": _("Invoice Total"),
```
vorasmit: not needed

PR #3224, opened 2025-03-20
refactor: Simplify financial quarter calculation with arithmetic

jeshani:
Replace if-elif with mathematical approach for quarter determination


vorasmit:
Hey, thanks for the contribution.

This reduces the readability and hence is less preferred. Would have considered this if it improved perf by any extent, but that's not the case.

PR #3255, opened 2025-04-02
refactor: update imports & switch to instance variables

khushi8112:
This PR updates the India Compliance app to align with the depreciation refactor in the Asset module. In the linked PR [Refactor PR Link](https://github.com/frappe/erpnext/pull/45251), depreciation-related functions that were previously part of the `asset_depreciation_schedule.py` file have been moved to separate classes. As a result, the imports have been updated accordingly.

Additionally, earlier, many functions relied on passing parameters explicitly. Now, since they are part of a class, instance variables are used instead of function arguments.

Note: The referenced refactor PR is not yet merged. This PR depends on it.

### DIFF
--- a/india_compliance/hooks.py
+++ b/india_compliance/hooks.py
@@ -360,7 +360,7 @@ regional_overrides = {
         "erpnext.accounts.party.get_regional_address_details": (
             "india_compliance.gst_india.overrides.transaction.update_party_details"
         ),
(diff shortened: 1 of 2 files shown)


khushi8112:
Hi @vorasmit 
Could you review this, please?

nabinhait (approved):

PR #3327, opened 2025-04-18
fix: uom for service item as "NA"

ljain112:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of unit of measurement (UOM) in HSN-wise summary reports to ensure correct display, especially for codes starting with "99".
	- Updated GSTR-1 report filter behavior for "HSN" business type to prevent unnecessary conditions.
- **Chores**
	- Removed obsolete code related to discontinued GSTR-1 report functionality.


vorasmit:
image

Looks like this is breaking the Excel Download.
When uploading excel, UOM NA is not supported.

vorasmit:
What was the issue faced here?

ljain112:
> What was the issue faced here?

Issue was in legacy report.

PR #3328, opened 2025-04-19
fix: cleaner bill number by truncating fy without special characters

vorasmit:


india_compliance/gst_india/doctype/purchase_reconciliation_tool/__init__.py
```diff
@@ -813,6 +813,12 @@ def reconcile_for_rule(self, purchases, inward_supplies, match_status, rules):
                     if not self.is_doc_matching(purchase, inward_supply, rules):
                         continue
 
+                    if match_status == MatchStatus.RESIDUAL_MATCH.value:
+                        if inward_supply.supplier_gstin == purchase.supplier_gstin:
+                            match_status = MatchStatus.SUGGESTED_MATCH.value
```
ljain112, line 818: @vorasmit 

Why was this added?

PR #3391, opened 2025-05-19
fix: don't add accumulated differences in last row

karm1000:
https://support.frappe.io/helpdesk/tickets/38211

## Summary by CodeRabbit

- **Refactor**
  - Improved the rounding process for GST data fields, ensuring each value is rounded to two decimal places without adjusting totals for rounding differences.


vorasmit:
Carryforward in #3399

PR #3437, opened 2025-06-09
fix: throw error if user is not permitted

aayush7908:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved permission checks for Bill Of Entry and GST Return Log features, ensuring users without the required access receive immediate error notifications.

### DIFF
--- a/india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.py
+++ b/india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.py
@@ -396,8 +396,8 @@ class BillofEntry(Document):
             frappe.msgprint(_("No Purchase Invoices selected"))
             return
 
-        frappe.has_permission("Bill Of Entry", "write")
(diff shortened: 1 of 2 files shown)


india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.py
```diff
@@ -396,8 +396,8 @@ def get_items_from_purchase_invoice(self, purchase_invoices):
             frappe.msgprint(_("No Purchase Invoices selected"))
             return
 
-        frappe.has_permission("Bill Of Entry", "write")
-        frappe.has_permission("Purchase Invoice", "read")
+        frappe.has_permission("Bill Of Entry", "write", throw=True)
```
vorasmit, line 399: @coderabbitai

It's important to throw here as without it, it would return `true` or `false`, which could then execute the code further. Please take a note of it and add this to review point.

PR #3455, opened 2025-06-15
fix: for "TypeError bad operand type for abs() 'NoneType'"

karanwilson:
After adding 12% CESS in the 28% tax tempate, when we mark CESS as 0% in the rest of the tax templates (5% , 12%) (except in the 0% tax template) we are getting the following error :-
[py_nontype_.err_msg.txt](https://github.com/user-attachments/files/20743967/py_nontype_.err_msg.txt)

Fix: Adding 0 when tax_rate is None.


ljain112:
Thank you for your contribution.
PR is continued in https://github.com/resilient-tech/india-compliance/pull/3461

PR #3755, opened 2025-10-27
fix: absolute item total value for rcm invoice

ljain112:
Issue: Error on Sales RCM Credit Note e-invoice generation.

image

Frappe Support Issue: https://support.frappe.io/helpdesk/tickets/51639?view=VIEW-HD+Ticket-771

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected e-invoice total value for reverse-charge transactions to use the absolute value of the taxable amount and apply rounding to two decimals, ensuring accurate and consistent invoice totals.

### DIFF
--- a/india_compliance/gst_india/utils/e_invoice.py
+++ b/india_compliance/gst_india/utils/e_invoice.py
@@ -685,7 +685,7 @@ class EInvoiceData(GSTTransactionData):
         )
 
         if self.doc.is_reverse_charge:
-            item_details["total_value"] = item.taxable_value
+            item_details["total_value"] = abs(self.rounded(item.taxable_value, 3))
 
         if batch_no := self.sanitize_value(
             item.batch_no, max_length=20, truncate=False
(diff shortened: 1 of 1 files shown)


india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -685,7 +685,7 @@ def update_item_details(self, item_details, item):
         )
 
         if self.doc.is_reverse_charge:
-            item_details["total_value"] = item.taxable_value
+            item_details["total_value"] = abs(self.rounded(item.taxable_value, 3))
```
karm1000:
```suggestion
            item_details["total_value"] = abs(self.rounded(item.taxable_value, 2))
```
precision should be 2 as the regex is `^\d+\.?\d{0,2}$`

ljain112: I tried to keep it consistent with above code.

"unit_rate": (
                    abs(self.rounded(item.taxable_value / item.qty, 3))

karm1000: > I tried to keep it consistent with above code.
> 
> "unit_rate": ( abs(self.rounded(item.taxable_value / item.qty, 3))

but wouldn't it throw the error from GST API? as it only accepts precision 2

PR #3782, opened 2025-11-07
fix: set taxes and totals before validating subcontracting transaction

### LINKED ISSUE #3777: Update grand total and taxes fields in stock entry in backend even if transaction is not eligible.

karm1000:
Fixes: #3777

## Summary by CodeRabbit

* **Refactor**
  * Tax initialization now runs earlier and only once, preventing duplicate calculations.

* **Bug Fixes / Tests**
  * total_taxes now shows 0.0 (not blank) for subcontracting orders with unregistered entities and for material receipt stock entries.

* **Tests**
  * Updated ITC-04 export expectations: adjusted an item's reported value from 0.0 to 200.0.

### DIFF
--- a/india_compliance/gst_india/overrides/subcontracting_transaction.py
+++ b/india_compliance/gst_india/overrides/subcontracting_transaction.py
@@ -178,6 +178,13 @@ def onload(doc, method=None):
 
 
 def validate(doc, method=None):
+    field_map = (
+        STOCK_ENTRY_FIELD_MAP
+        if doc.doctype == "Stock Entry"
+        else SUBCONTRACTING_ORDER_RECEIPT_FIELD_MAP
+    )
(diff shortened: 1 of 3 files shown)


india_compliance/gst_india/overrides/subcontracting_transaction.py
```diff
@@ -177,6 +177,15 @@ def onload(doc, method=None):
         doc.set_onload("e_waybill_info", e_waybill_info)
 
 
+def before_validate(doc, method=None):
```
ljain112: I think it should be in validate only because erpnext also calculates everything in validate, which can cause inconsistency.

ljain112:
@karm1000 resolve test cases

ljain112:
@mergifyio backport version-15-hotfix

PR #3987, opened 2026-02-13
fix: add support for 8-digit HSN code validation

ljain112:
Issue: For some companies, it is a requirement to use 8 digit HSN Code.


karm1000 (approved):
LGTM

PR #4003, opened 2026-02-19
chore: update release branches to version-16 for release config

karm1000:


sagarvora:
14, 15, and 16? we want to maintain all 3? or at least 15 & 16? -> should be done in discussion with smit

PR #4023, opened 2026-02-25
fix: update default base branch reference in server tests workflow

### LINKED ISSUE #4022: DX: Hardcore base branch of frappe and erpnext in test
- Reason:  https://github.com/resilient-tech/india-compliance/actions/runs/22350277299/job/64675307827?pr=4018

Image

- Need to make changes in all branches differently

Abdeali099:
- closes: #4022

### DIFF
--- a/.github/workflows/server-tests.yml
+++ b/.github/workflows/server-tests.yml
@@ -30,7 +30,7 @@ on:
             - "**.html"
             - "**.csv"
 env:
-    BRANCH: ${{ inputs.base_ref || github.base_ref || github.ref_name }}
+    BRANCH: ${{ inputs.base_ref || 'develop' }}
     APP_NAME: ${{ inputs.app_name || 'india_compliance' }}
 
 jobs:


Abdeali099:
need to discuss with @sagarvora

.github/workflows/server-tests.yml
```diff
@@ -30,7 +30,7 @@ on:
             - "**.html"
             - "**.csv"
 env:
-    BRANCH: ${{ inputs.base_ref || github.base_ref || github.ref_name }}
+    BRANCH: ${{ inputs.base_ref || 'develop' }}
```
karm1000, line 33:
```suggestion
    BRANCH: ${{ github.event.inputs.branch || github.base_ref || 'develop' }}
```

Maybe this?

Abdeali099: `github.base_ref ` is problematic, 

It will take base branch as input example `A`, but `A` not exist in frappe. 

and `github.event.inputs.branch` is not useful because we are not defining `workflow_dispatch` in `on` section for manual trigger.

But I think we can use it by changing main section as ERPNext.

https://github.com/frappe/erpnext/blob/40a5b14d642f31a6d2302b871b9e3f9434741ef5/.github/workflows/server-tests-mariadb.yml#L19

PR #4028, opened 2026-02-25
fix: correct taxable amount when group by invoice in GST Purchase Register

ljain112:
Manual Bakcport: https://github.com/resilient-tech/india-compliance/pull/4027


karm1000:
This should be backported by this https://github.com/resilient-tech/india-compliance/pull/4027

PR #4077, opened 2026-03-11
fix: patch to update description of `enable_audit_trail`

diptanilsaha:
Ref: #4071

### DIFF
--- a/india_compliance/patches.txt
+++ b/india_compliance/patches.txt
@@ -78,3 +78,4 @@ india_compliance.patches.v16.remove_legacy_report_fixtures
 india_compliance.patches.v15.unset_auth_token
 india_compliance.patches.v15.update_itc_claim_period
 india_compliance.patches.v15.migrate_logo_for_printing
+execute:from india_compliance.audit_trail.setup import create_custom_fields; create_custom_fields()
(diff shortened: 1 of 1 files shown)


india_compliance/patches.txt
```diff
@@ -21,7 +21,6 @@ execute:from india_compliance.audit_trail.setup import setup_fixtures; setup_fix
 india_compliance.patches.v14.set_default_for_audit_trail_notification
 india_compliance.patches.post_install.update_state_name_to_puducherry
 india_compliance.patches.post_install.rename_import_of_capital_goods
-execute:from india_compliance.audit_trail.setup import create_custom_fields, CUSTOM_FIELDS; create_custom_fields(CUSTOM_FIELDS)
```
ljain112, line 24: You can just increase the patch counter

`execute:from india_compliance.audit_trail.setup import create_custom_fields, CUSTOM_FIELDS; create_custom_fields(CUSTOM_FIELDS)  #1`

diptanilsaha: `create_custom_fields` doesn't have any parameters. Ref: https://github.com/resilient-tech/india-compliance/blob/develop/india_compliance/audit_trail/setup.py#L21

PR #4108, opened 2026-03-18
ci: add workflow_dispatch

shubhdoshi21:
## Add `workflow_dispatch` with Custom Frappe/ERPNext Remotes and Concurrency control

Adds a manual `workflow_dispatch` trigger to `server-tests.yml` so the CI can be run against custom forks or branches of Frappe and ERPNext without modifying any files. Also fixes `install.sh` to correctly handle ERPNext forks that have a different app/repo name than `erpnext`.

Adds `concurrency` group keyed by event type + PR number (or run ID for dispatches), with `cancel-in-progress: true` to avoid redundant runs on the same PR.

### DIFF
--- a/.github/helper/install.sh
+++ b/.github/helper/install.sh
@@ -27,7 +27,10 @@ wkpid=$!
 
 pip install frappe-bench
 
-git clone "https://github.com/frappe/frappe" --branch "$BRANCH_TO_CLONE" --depth 1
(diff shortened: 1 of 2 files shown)


karm1000:
@shubhdoshi21 resolve conflicts

karm1000 (approved):
LGTM

PR #4112, opened 2026-03-19
fix: failing test

Priyal208:
Any ERPNext Test Utilities should be imported with unittest mock to avoid repeated creation of BootStrapTestData of ERPNext and avoid breaking of test after refactor of ERPNext as in: https://github.com/frappe/erpnext/commit/518800cd2f5e7fef00e70a5c90da14af2a5e522c and https://github.com/frappe/erpnext/commit/5a4a77f5d27510ce809923b81028a8f59b3a9dba

### DIFF
--- a/india_compliance/gst_india/doctype/bill_of_entry/test_bill_of_entry.py
+++ b/india_compliance/gst_india/doctype/bill_of_entry/test_bill_of_entry.py
@@ -3,11 +3,11 @@
 
 import json
 import re
+import unittest.mock as mock
 
 import frappe
 from frappe.tests import IntegrationTestCase
 from frappe.utils import today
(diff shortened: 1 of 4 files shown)


india_compliance/gst_india/overrides/party.py
```diff
@@ -15,6 +15,9 @@
 
 
 def validate_party(doc, method=None):
+    if frappe.flags.in_test_setup:
```
vorasmit: anti-pattern

Priyal208:
@greptileai

Priyal208:
@greptileai

ljain112:
@mergifyio backport version-16-hotfix

PR #4125, opened 2026-03-25
fix: ensure company GSTIN options are set only for new GSTR 3B reports

ljain112:
Issue: Due to autosetting of GSTIN on onload.A user with multiple GSTINs in a single company was not able to re-save because the company GSTIN can be set only once.

image


ljain112:
@greptileai

karm1000 (approved):

PR #4131, opened 2026-03-26
fix: guard dialog onchange handlers against early calls during construction

ljain112:
regression: https://github.com/frappe/frappe/pull/37384
Issue: Previously, field defaults were applied via set_input(default_value), which only updates the DOM silently. After this PR, defaults are applied via set_value(default_value), which triggers onchange.
This causes onchange to fire for all fields with a default during new frappe.ui.Dialog() construction — before the dialog has been initialised.

Frappe Support Issue: https://support.frappe.io/desk/hd-ticket/63019

### DIFF
--- a/india_compliance/gst_india/client_scripts/e_waybill_actions.js
+++ b/india_compliance/gst_india/client_scripts/e_waybill_actions.js
@@ -1326,7 +1326,7 @@ async function update_gst_tranporter_id(dialog) {
             transporter,
             "gst_transporter_id",
         );
-        transporter_id = response?.gst_transporter_id;
+        transporter_id = response?.gst_transporter_id || "";
     }
 
     dialog.set_value("gst_transporter_id", transporter_id);
--- a/india_compliance/gst_india/client_scripts/e_waybill_actions.js
+++ b/india_compliance/gst_india/client_scripts/e_waybill_actions.js
@@ -1318,14 +1318,18 @@ function is_e_waybill_cancellable(frm) {
 }
 
 async function update_gst_tranporter_id(dialog) {
+    let transporter_id = "";
     const transporter = dialog.get_value("transporter");
-    const { message: response } = await frappe.db.get_value(
-        "Supplier",
-        transporter,
-        "gst_transporter_id",
-    );
+    if (transporter) {
+        const { message: response } = await frappe.db.get_value(
+            "Supplier",
+            transporter,
+            "gst_transporter_id",
+        );
+        transporter_id = response?.gst_transporter_id;
+    }
 
-    dialog.set_value("gst_transporter_id", response.gst_transporter_id);
+    dialog.set_value("gst_transporter_id", transporter_id);
 }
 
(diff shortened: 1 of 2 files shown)


india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.js
```diff
@@ -1006,7 +1006,7 @@ class ImportDialog {
     }
 
     async fetch_import_history() {
-        if (!this.company_gstin) return;
+        if (!this.company_gstin || !this.return_type || !this.date_range) return;
```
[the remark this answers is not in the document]
ljain112: It will still work because in backend it will fallback to current date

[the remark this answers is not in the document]
ljain112: It will still work because in backend it will fallback to current date

ljain112:
@greptileai
P2 Consider validating date_range array contents - It will still work because in backend it will fallback to current date

ljain112:
@greptileai

india_compliance/gst_india/client_scripts/e_waybill_actions.js
```diff
@@ -1318,14 +1318,18 @@ function is_e_waybill_cancellable(frm) {
 }
 
 async function update_gst_tranporter_id(dialog) {
+    let transporter_id = "";
```
karm1000, line 1321: making this as `null` instead of empty string would be better

india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.js
```diff
@@ -1119,6 +1119,14 @@ class ImportDialog {
                 default: this.frm.doc.inward_supply_period,
                 onchange: async () => {
                     const period = this.dialog.get_value("period");
+                    if (!period) return;
+
+                    if (period === "Custom") {
```
vorasmit: refactor.

india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.js
```diff
@@ -1119,6 +1119,14 @@ class ImportDialog {
                 default: this.frm.doc.inward_supply_period,
```
karm1000, line 1119: At the initialisation of dialog `init_dialog`. Two API calls for `get_import_history` are being made for the same arguments. 

Reason: One by one company_gstin, return_type, date_range, period are being set in sequence. When the fields till date_range are set, the condition results in `false` and it makes API call. After that `period` is being set so onchange of `period` results in calling the same API.

ljain112: The issue is in frappe we will check if Frappe is fixing it or not.

PR #4142, opened 2026-03-27
fix: use "is_boe_applicable" for checking valid boe purchase invoice

### LINKED ISSUE #4133: Refactor in bill of entry
<a href="#">P1</a> **Missing `is_boe_applicable` guard in `validate_purchase_invoices`**

The validation correctly rejects invoices whose `gst_category` is not in `IMPORT_GST_CATEGORIES`. However, a SEZ goods invoice that carries GST taxes (IGST charged directly by the SEZ unit) has `gst_category = "SEZ"` — which passes this check — but its `is_boe_applicable` is `0` because the ITC is settled directly through the PI, not a BOE.

If a user manually navigates to the Bill of Entry doctype and links such an invoice, the `gst_category` check passes silently. The BOE would have no items (since `get_pi_items` filters on `is_boe_applicable == 1` and `pending_boe_qty > 0`) and would be an empty document.

Adding an explicit guard here makes the server-side validation self-consistent with the PI-level filtering:

```python
if not invoice.is_boe_applicable:
    frappe.throw(
        _(
            "Bill of Entry is not applicable for Purchase Invoice {0}"
        ).format(invoice.name)
    )
```

_Originally posted by @greptile-apps[bot] in https://github.com/resilient-tech/india-compliance/pull/4103#discussion_r2995067188_

ljain112:
closes: https://github.com/resilient-tech/india-compliance/issues/4133

### DIFF
--- a/india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.py
+++ b/india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.py
@@ -154,7 +154,14 @@ class BillofEntry(Document):
         purchase_invoices = frappe.get_all(
             "Purchase Invoice",
             filters={"name": ["in", pi_names]},
-            fields=["docstatus", "gst_category", "name", "company", "company_gstin"],
+            fields=[
+                "docstatus",
+                "gst_category",
+                "name",
+                "company",
+                "company_gstin",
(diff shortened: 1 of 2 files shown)


ljain112:
@greptileai

Issue #4054, opened 2026-03-07
Incorrect taxes in mapped doc.


ljain112:
In multiple transactions, like mapped doc, new doc, where the address is updated in erpnext, the correct place of supply, taxes are not correct, in set_missing_details, only fields which are empty are updated.

Fraape Supporty Issue:
-  https://support.frappe.io/desk/hd-ticket/61925
- https://support.frappe.io/desk/hd-ticket/51240

PR #4223, opened 2026-04-17
fix: update place of_supply and address if erpnext updates party address on new doc

ljain112:
Issue: If the party address is empty in the doc, erpnext updates the party address, but the place of supply remains unchanged.

Frappe Support Issue: https://support.frappe.io/desk/hd-ticket/51240
ref issue: #4054

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -900,7 +900,7 @@ def _get_address_fields(doctype, party_details=None):
         )
 
     elif doctype == "Stock Entry":
-        if party_details.get("is_inward_stock_entry"):
+        if party_details and party_details.get("is_inward_stock_entry"):
             address_fields.update(
                 company_gstin_field="bill_to_gstin",
                 party_gstin_field="bill_from_gstin",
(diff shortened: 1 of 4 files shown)


ljain112:
@greptileai

ljain112:
@coderabbitai review

india_compliance/gst_india/overrides/transaction.py
```diff
@@ -1604,6 +1666,10 @@ def validate_transaction(doc, method=None):
     validate_item_wise_tax_detail(doc)
 
 
+def _is_sales_transaction(doctype):
```
vorasmit, line 1669: not what it means. could cause confusion at places where it's used.

india_compliance/hooks.py
```diff
@@ -141,6 +141,7 @@
             "india_compliance.gst_india.overrides.transaction.onload",
         ],
         "before_print": "india_compliance.gst_india.overrides.transaction.before_print",
+        "before_validate": "india_compliance.gst_india.overrides.transaction.before_validate_transaction",
```
vorasmit, line 144: UI / frontend issue is being resolved with backend hooks.
Can we use JS hooks instead and trigger what happens on change of address?

ljain112: This is not ui issue, basically, the party doesn't triggered  and data is being set by erpnext in the backend

india_compliance/gst_india/overrides/transaction.py
ljain112: Reverting it.

india_compliance/gst_india/overrides/transaction.py
```diff
@@ -1540,11 +1566,47 @@ def before_validate_transaction(doc, method=None):
...
+    """
+
+    if not doc.get("_party_address_not_set"):
+        return
+
+    address_field = _get_address_fields(doc.doctype).get("party_address_field")
+    if not doc.get(address_field):
+        return
+
+    gst_details = get_gst_details(doc.as_dict(), doc.doctype, doc.company, update_place_of_supply=True)
```
vorasmit, line 1591: always update.

vorasmit (approved):

PR #4239, opened 2026-04-23
fix: remove conditional check for transported GSTIN for extend e-waybill

ljain112:
there is no reliable way to determine upfront whether the user is actually eligible to extend.

GST Transporter ID set may not necessarily be valid GST Transporter ID.
Only if Valid GST transporter ID is set, user cannot extend.

Keeping things simpler, by letting GST portal validate this after API call is made.

---

This is for both workflows:
- Immediately extend
- Schedule

As after user has scheduled, transporter GSTIN can change.


vorasmit (approved):

vorasmit:
Error handling for error raised if user is not able to extend validity.

PR #4302, opened 2026-05-14
fix: reset gst details on cross doctype mapping

ljain112:
Issue: hen creating a Purchase Order (PO) from a Sales Order (SO), the PO incorrectly inherits the place_of_supply from the SO. While this reflects the customer's state on the SO, the PO should instead determine this field based on the supplier's address or GSTIN. This mismatch causes India Compliance's GST validation to throw an error (e.g., "Invalid GST Account — Row #2: Cannot charge CGST/SGST for inter-state supplies"), blocking the workflow entirely.


ljain112:
@greptileai

karm1000:
This is working as expected. 
@vorasmit Does this PR need any further changes?

vorasmit (approved):

PR #4345, opened 2026-06-01
fix: validate credentials only when relevant fields are updated

### LINKED ISSUE #55366: IC app Please set credential... pop up message issue
### Information about bug

The message below pops up after clearing the demo data (deleting) and after creating tax templates from the company, both actions are getting successful, but this pop up messsage confuces user and is not required, so it should not come.

Image

Image

### Module

accounts

### Version

v16

### Installation method

None

```shell

```

ljain112:
closes: https://github.com/frappe/erpnext/issues/55366


ljain112:
@greptileai

vorasmit (approved):

PR #4353, opened 2026-06-02
test: update get_materials_from_supplier calls to use frappe.flags.args for subcontracting details

ljain112:
Related PR: https://github.com/frappe/erpnext/pull/55537


ljain112:
@greptileai

PR #4361, opened 2026-06-03
fix: set gst details and treatment in validate instead of before_save

ljain112:
- gst_treatment and item_gst_details were updating in after_save and after_submit
- moved it to the validate hook
- As per the erpnext convention, validation should happen in validate
- Item GST Treatment will get reset


ljain112:
@mergifyio rebase

ljain112:
@greptileai

ljain112:
@greptileai review

PR #4362, opened 2026-06-03
fix: sum rounded total for correct e-waybill total

ljain112:
Issue: The sum of item-wise taxable value is not equal to the total taxable value.
The issue occurs only for the e-Waybill JSON upload.
image

Frappe Support Issue: https://support.frappe.io/helpdesk/tickets/67983


ljain112:
@greptileai

vorasmit (approved):

PR #4374, opened 2026-06-04
fix: update import paths for erpnext functions

ljain112:
related pr: 
- https://github.com/frappe/erpnext/pull/55596
- https://github.com/frappe/erpnext/pull/55327


ljain112:
@greptileai

PR #4432, opened 2026-06-17
fix: implement taxable value update on stock entry submission

ljain112:
Issue: On submit, ERPNext recomputes the outgoing valuation (basic_rate → amount) with the actual FIFO/batch rate (update_rate_on_stock_entry / update_rate_on_subcontracting_receipt) on a freshly loaded doc via db_update — bypassing the validate doc_event where India Compliance sets taxable_value and base_grand_total. Those GST fields stay frozen at their draft-time estimate, so submitted Stock Entries and Subcontracting Receipts end up with base_grand_total ≠ total_amount and per-row taxable_value ≠ amount, feeding wrong values into e-Waybill / ITC-04.

Frappe Support Issue: https://support.frappe.io/desk/hd-ticket/63655

Note: Unable to reproduce the issue in test.


ljain112:
Issue only at repost and not changing value on repost.

PR #4433, opened 2026-06-17
fix: add fuzzy match rule without date and corresponding test case

ljain112:
Issue: When a supplier had two invoices with the same GST amount and their bill numbers were stored in a slightly different format in GSTR-2A vs ERP, the tool could cross-match them — e.g. showing QKI/055190 (2A) as a "Suggested Match" to QKI/053612 (ERP), while the genuine same-number invoices were left as Missing in PI / Missing in 2A/2B.

Root cause: fuzzy_match applied the 10-day bill-date check before comparing the (normalized) bill numbers. For the same invoice booked with a >10 day date gap, every bill-number rule was skipped, so the pair fell through to the amount-only Residual Match rule, which paired invoices purely by tax amount and was then promoted to "Suggested Match".

Fix: Give the bill number preference over an amount-only match. Added a FUZZY_MATCH_WITHOUT_DATE rule (bill no matches, date ignored) to GSTIN_RULES, placed after the date-constrained bill rules and before the residual rule. Matching priority is now: bill + date → bill no → date + amount.

Frappe Support Issue: https://support.frappe.io/helpdesk/tickets/67857?view=VIEW-HD+Ticket-771

### DIFF
--- a/india_compliance/gst_india/doctype/purchase_reconciliation_tool/__init__.py
+++ b/india_compliance/gst_india/doctype/purchase_reconciliation_tool/__init__.py
@@ -40,7 +40,7 @@ class Fields(Enum):
 class Rule(Enum):
     EXACT_MATCH = "Exact Match"
     FUZZY_MATCH = "Fuzzy Match"
-    FUZZY_MATCH_WITHOUT_DATE = "Fuzzy Match Without Date"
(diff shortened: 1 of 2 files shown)


ljain112:
Issue was dur to residual mismatch being called as suggested match invalid fix.

PR #4478, opened 2026-06-26
fix: avoid duplicate get_import_history calls on dialog init

### LINKED ISSUE #4137: At the initialisation of dialog `init_dialog`. Two API calls for `get_import_history` are being made for the same arguments.
At the initialisation of dialog `init_dialog`. Two API calls for `get_import_history` are being made for the same arguments. 

Reason: One by one company_gstin, return_type, date_range, period are being set in sequence. When the fields till date_range are set, the condition results in `false` and it makes API call. After that `period` is being set so onchange of `period` results in calling the same API.

_Originally posted by @karm1000 in https://github.com/resilient-tech/india-compliance/pull/4131#discussion_r2994716987_

vorasmit:
## Summary

Closes #4137

When the **Import** dialog (`ImportDialog`) in the Purchase Reconciliation Tool initialises, it applies defaults to its fields (`return_type`, `company_gstin`, `period`, `date_range`). Each field's `onchange` handler calls `fetch_import_history()`, and `init_dialog()` also calls it explicitly. The early-return guard (`!company_gstin || !return_type || !date_range`) only blocks calls until all three are set — so once `period`/`date_range` defaults land, **two** `get_import_history` requests fire for the *same* arguments.

## Fix
`fetch_import_history()` now builds a signature from its arguments (`company_gstin`, `return_type`, `date_range`, `for_download`) and skips the request when it matches the previous one. A genuine field change produces a different signature, so user-driven refreshes still fetch as before.

## Why this approach
The `onchange` handlers are needed for interactive use, and the explicit init call guarantees one fetch regardless of frappe's default-application order. Deduping by argument signature fixes the duplicate robustly without depending on the exact order in which defaults fire.

## Testing
- Prettier passes on the changed file.
- Recommend a quick manual check: open the Download/Upload dialog and confirm only **one** `get_import_history` call is made on open (Network tab), and that changing GSTIN/return type/period still triggers a fresh call.

---

### DIFF
--- a/india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.js
+++ b/india_compliance/gst_india/doctype/purchase_reconciliation_tool/purchase_reconciliation_tool.js
@@ -935,23 +935,39 @@ class ImportDialog {
     async fetch_import_history() {
         if (!this.company_gstin || !this.return_type || !this.date_range) return;
 
-        // Skip duplicate requests for identical arguments
+        // Skip duplicate requests for identical arguments, whether already
+        // in-flight or last completed successfully. The signature is only
+        // marked as completed after a successful fetch, so a failed request
+        // can still be retried with the same arguments.
         const args_signature = JSON.stringify([
             this.company_gstin,
             this.return_type,
             this.date_range,
             this.for_download,
         ]);
-        if (args_signature === this._last_import_history_signature) return;
-        this._last_import_history_signature = args_signature;
+        if (
+            args_signature === this._last_import_history_signature ||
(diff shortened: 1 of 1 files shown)


vorasmit:
Good catch — fixed in eafce78f.

The signature was marked as seen *before* awaiting `get_import_history`, so a failed request left it cached and blocked retries with the same arguments. Now there are two signatures: a `_pending` one set before the call (so the init burst is still deduped while a request is in flight) and `_last` which is only set **after a successful fetch**, with `_pending` cleared in a `finally`. A failed fetch therefore no longer poisons the cache and can be retried.

---

Abdeali099 (approved):

PR #4479, opened 2026-06-26
ci: don't run server tests twice for releases

### LINKED ISSUE #4206: Two server test event are executed for release (push and pull_request)

vorasmit:
## Summary

Closes #4206

`initiate_release.yml` opens a PR from `version-X-hotfix` → `version-X` for each version (14/15/16). That PR runs the server tests via the **`pull_request`** event, and then merging it pushes to `version-X`, which was also in the **`push`** branch list of `server-tests.yml` — so the full suite ran **twice per version** on every release.

The existing `concurrency` group is keyed by `github.event_name`, so the `push` and `pull_request` runs land in different groups and aren't deduped that way.

## Fix
Remove the release-target branches (`version-14`, `version-15`, `version-16`) from the `push` trigger. Code only reaches those branches through the automated release PR, which already validates it pre-merge.

After this change, per release version:
- Release PR (`version-X-hotfix` → `version-X`): tested once via `pull_request`. ✅ (pre-merge gate)
- Post-merge push to `version-X`: no longer re-runs the suite.
- `on_release.yml` still runs on the `version-X` push independently (unchanged).
- `develop` and the `version-X-hotfix` branches keep their push-triggered runs.

## Alternative considered
The mirror image — keep the `push` runs and scope `pull_request` to exclude version-branch bases — also halves the runs, but that would make the release test post-merge (non-gating). Gating the release on the PR seemed preferable. Happy to flip it if you'd rather keep the push run as the release gate.

---

### DIFF
--- a/.github/workflows/server-tests.yml
+++ b/.github/workflows/server-tests.yml
@@ -14,16 +14,18 @@ on:
 
   pull_request:
 
+  # Note: the release-target branches (version-14/15/16) are intentionally
+  # excluded here. Code reaches them only via the automated release PR
+  # (version-X-hotfix -> version-X), which is already tested by the
+  # pull_request event above. Including them would run the suite a second
+  # time on the post-merge push. See #4206.
   push:
     branches:
       [
         develop,
         version-14-hotfix,
-        version-14,
         version-15-hotfix,
-        version-15,
         version-16-hotfix,
-        version-16,
       ]
 
   workflow_dispatch:


vorasmit:
Good catch on the coverage side-effect (the bot's P2). Some context on the trade-off:

The `tests` job still runs for the release PR via `pull_request`, but the `coverage` "Wrap Up" job is gated to `github.event_name == 'push'`, so with `version-14/15/16` out of the push triggers, Codecov won't get an upload for those branches on the release path.

I'd lean toward **accepting the gap**, for two reasons:
1. The release PR merges `version-X-hotfix → version-X` with identical content, and `version-X-hotfix` **is** still in the push list — so the coverage for what gets released is already uploaded under the hotfix branch. Only the per-branch `version-X` view goes stale, not the underlying data.
2. This approach keeps the test run as a **pre-merge gate** on the release PR, which seemed more valuable for release integrity than a post-merge coverage upload.

The alternative — keep the `push` runs and instead scope `pull_request` to exclude version-branch bases — would preserve coverage uploads exactly as today, but then the release is validated *post-merge* rather than gated. Happy to switch to that if you'd rather keep coverage continuity over gating; both fix the double-run. Let me know which you prefer.

---

vorasmit:
Updated coverage is important -> user viewing the repo for specific brach would see stale coverage.
Running server tests before merge is also important.

It's a small compromise to run them twice. We add comment to the issue, and close this PR and the related issue.

vorasmit:
Closing per the decision on #4206 — keeping both up-to-date per-branch coverage and pre-merge test gating is worth the minor cost of running the suite twice on release. Thanks for the review!

---

PR #4482, opened 2026-06-26
feat(gst): allow marking transactions Out of Scope of GST

### LINKED ISSUE #1455: Flag transactions as out of scope of GST => Enforce GST on all other transactions
- check box for document
- Filters in e-Waybill and e-Invoice applicability
- Update in Item GST Treatment
- Filters in Purchase reco tool
- Validate GST Accounts

vorasmit:
> **no-docs** — end-user documentation for this override will follow as a separate PR in `india-compliance-docs`.

## Summary

Adds a manual document-level override **"Out of Scope of GST"** so a transaction (e.g. merchant trade) is kept entirely outside the GST framework — no GST accounts, no e-Waybill/e-Invoice, and excluded from every GST report and return. Self-billing / internal-transfer invoices (which the framework already treats as outside GST) are auto-flagged via a migration patch.

Closes #1455

## What's included (maps to the issue checklist)

- **Checkbox on documents** — new `is_out_of_scope_of_gst` Check on all sales/purchase transactions (same doctype set as reverse charge; POS Invoice & Stock Entry excluded, matching existing convention).
- **Item GST Treatment** — new item `gst_treatment` value `"Out of Scope"`, assigned to all items when the flag is set.
- **Validate GST Accounts** — the flag is routed through `ignore_gst_validations` (skips the GST pipeline) and a dedicated `validate_out_of_scope()` forbids any GST tax account on the document, detected via `gst_tax_type` (`set_gst_tax_type` resolves it for out-of-scope docs). Wired into the create path **and** the after-submit hooks.
- **Reporting exclusion** — one WHERE on the outward backbone (`GSTR1Query.get_base_query`) and one on the inward backbone (`GSTR3BInwardQuery.get_base_purchase_query`) cover GSTR-1/3B + the sales/purchase/HSN registers. Explicit filters added to the tax-rate-wise, ITC-availed, account-wise (inside the existing JE/BOE guard) and e-invoice-summary reports. Flagged docs surface in a dedicated **"Excluded from Report (Out of Scope of GST)"** bucket in the Document Issued Summary, so issued-document counts stay correct.
- **e-Waybill & e-Invoice applicability** — predicate guards force "Not Applicable", plus a hard block in `EWaybillData.validate_applicability` (the manual-generation path the predicate doesn't gate).
- **Purchase Reconciliation** — `set_reconciliation_status` is flag-aware and now runs on the out-of-scope early-return path, so flagged Purchase Invoices become "Not Applicable" and drop out of both the Purchase Reconciliation Tool and IMS.
- **Auto-enable patch** — `v16/set_out_of_scope_for_internal_transactions` backfills the flag for historical self-billing **invoices** (`company_gstin == party_gstin`), scoped to Indian companies, with a both-empty-GSTIN guard and idempotency. Delivery Notes/Receipts are intentionally excluded (same-GSTIN goods movements still need an e-Waybill). The `create_custom_fields` patch marker is bumped so existing sites create the new field on migrate.

## Design notes

- **Document-level checkbox is authoritative**; the new item treatment value is set from it. Reporting exclusion happens at the query level, so a Non-GST-style "still reported" leak is avoided.
- `"Out of Scope"` is deliberately **not** in `TAXABLE_GST_TREATMENTS`, so it is treated as non-taxable everywhere automatically.
- No `no_copy` on the field — it carries through amend and the document flow (SO → DN → SI), matching `is_reverse_charge`.
- No new GST Settings toggle (checkbox ships always-on, per maintainer direction).

## Tests

Added to the parametrized `TestTransaction`: items become `"Out of Scope"`, charging a GST account throws (insert and after-submit paths), and an out-of-scope Purchase Invoice gets `reconciliation_status = "Not Applicable"`.

## Verification

`py_compile`, `node --check`, `ruff check`, `ruff format --check` and `prettier --check` all pass on the changed files. The Frappe integration suite was **not** run in the authoring environment (no `bench`/DB) — recommend `bench run-tests --module india_compliance.gst_india.overrides.test_transaction` on a real site before merge.

https://claude.ai/code/session_01H3HCYgtY9m57W2Mm7tqdTu

### DIFF
--- a/india_compliance/gst_india/overrides/transaction.py
+++ b/india_compliance/gst_india/overrides/transaction.py
@@ -836,7 +836,8 @@ def get_gst_details(
 
     # Taxes Not Applicable
     if (
-        (
+        party_details.get("is_out_of_scope_of_gst")
+        or (
             not allow_same_gstin
             and (
                 party_details.get(company_gstin_field)
@@ -1391,6 +1392,10 @@ class ItemGSTTreatment:
         self.doc = doc
         is_sales_transaction = doc.doctype in SALES_DOCTYPES
 
+        if doc.get("is_out_of_scope_of_gst"):
+            self.set_for_out_of_scope()
+            return
+
         if is_sales_transaction and is_overseas_doc(doc):
             self.set_for_overseas()
             return
@@ -1411,6 +1416,10 @@ class ItemGSTTreatment:
 
         self.update_gst_treatment_map()
         self.set_default_treatment()
+
+    def set_for_out_of_scope(self):
+        for item in self.doc.items:
+            item.gst_treatment = "Out of Scope"
 
     def set_for_overseas(self):
         for item in self.doc.items:
@@ -1612,6 +1621,9 @@ def _update_place_of_supply_and_taxes(doc):
 
 def validate_transaction(doc, method=None):
     if ignore_gst_validations(doc):
+        if doc.get("is_out_of_scope_of_gst"):
+            validate_out_of_scope(doc)
+
         set_gst_tax_type(doc)
         update_item_gst_treatment(doc)
         update_item_gst_details(doc)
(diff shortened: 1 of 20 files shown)


india_compliance/gst_india/overrides/transaction.py
```diff
@@ -1834,10 +1846,27 @@ def reset_gst_details_on_cross_mapping(target_doc, source_doc):
...
+        or doc.get("is_opening") == "Yes"
+        or doc.get("is_out_of_scope_of_gst")
+    ):
         return True
 
 
+def validate_out_of_scope(doc):
+    # no GST account allowed when transaction is out of scope of GST
+    # gst_tax_type is blank under the skip gate, so match account heads directly
+    gst_accounts = set(get_all_gst_accounts(doc.company))
```
ljain112: This can be directly done using gst_tax_type we are already setting it. Either remove that.

vorasmit: Done in df66694 — `validate_out_of_scope` now checks `row.gst_tax_type` instead of matching account heads, and the `get_all_gst_accounts` lookup is dropped.

One adjustment was needed: `set_gst_tax_type` builds an empty account→tax-type map under `ignore_gst_validations` (which now includes out-of-scope), so `gst_tax_type` was coming back blank. I made `set_gst_tax_type` still resolve tax types for out-of-scope docs (the one ignore-case that must detect GST accounts), and `validate_out_of_scope` calls it before checking `row.gst_tax_type`.

---

PR #4485, opened 2026-06-26
refactor: drop Tax Category Source State (gst_state)

### LINKED ISSUE #4480: This field (`gst_state`) can be scrapped in the following PR
This field (`gst_state`) can be scrapped in the following PR

_Originally posted by @vorasmit in https://github.com/resilient-tech/india-compliance/pull/4434#discussion_r3464697166_

vorasmit:
## Summary

Closes #4480 (follow-up to #4434)

Users should no longer have to configure a *Source State* (`gst_state`) on a Tax Category to get the right tax template. With the `is_india_compliance_default` flag from #4434, template auto-selection is fully determined by the transaction's **place of supply** (Inter State / Reverse Charge, derived from the address at transaction time). This PR removes `gst_state` and all of its usage.

> Per the discussion on #4480, this targets **develop (v17) only** — `gst_state` is retained on v15/v16.

## Changes
- **Field removal** — drop the `gst_state` ("Source State") field and its now-orphaned `tax_category_column_break` from the Tax Category custom fields (`custom_fields.py`); remove `gst_state` from the shipped `tax_defaults.json`.
- **Selection logic** — `get_tax_template()` now selects purely on `is_india_compliance_default` + `is_inter_state`/`is_reverse_charge`; the `state_code` argument and `STATE_NUMBERS` lookup are gone. Both call sites (`get_gst_details`, `set_reverse_charge`) updated. (The loop is kept so a stray manually-`db_set` duplicate can't break selection.)
- **Validation** — `tax_category.validate` drops the `gst_state` branch, keeping only the duplicate-default guard.
- **Client script** — the Tax Category client script existed only to toggle `gst_state` ⇄ `is_india_compliance_default`; removed it and its `doctype_js` hook.
- **Tests** — removed the state-specific selection/validation tests (and the now-unused `SOURCE_STATE`); default, inter-state, reverse-charge, explicit-category and duplicate-default tests remain.
- **Migration** — new `v17/remove_tax_category_source_state` patch deletes the `Tax Category-gst_state` and column-break custom fields. **Address** `gst_state`/`gst_state_number` are untouched.

## Migration impact
Existing user Tax Categories survive as plain categories — still usable when explicitly set on a party/transaction — they are simply no longer state-specific. The shipped defaults already covered all Inter State / Reverse Charge combinations, so default selection is unchanged.

## Testing
- `ruff` clean; Python parses; `tax_defaults.json` valid.
- Server tests (incl. `test_tax_category.py`) require a bench and will run in CI.

---


vorasmit:
Thanks @coderabbitai — addressed one, and I think the other doesn't apply here:

**✅ Disabled-default uniqueness (`tax_category.py`)** — good catch, fixed in c63129c. `get_tax_template` filters `disabled = 0`, but the uniqueness check was counting disabled defaults as conflicts, so disabling a default would block creating a replacement. The validator now skips disabled docs and only treats **active** defaults as conflicts.

**⏭️ Migration reconciliation of "legacy default categories" (patch)** — I don't think this scenario can occur. The premise is that rows can have `is_india_compliance_default = 1` together with a `gst_state`, but #4434 prevents exactly that:
- `tax_category.validate` throws on `is_india_compliance_default AND gst_state` ("…cannot have a Source State"), so no such row can be saved.
- the `set_india_compliance_default_tax_category` backfill only flags categories where `gst_state is not set`, one per `(is_inter_state, is_reverse_charge)` scenario.

So every `is_india_compliance_default = 1` row already has an empty `gst_state`, and there is at most one active default per combination. Removing `gst_state` therefore doesn't introduce any new ambiguity — selection by `(is_inter_state, is_reverse_charge, is_india_compliance_default, disabled=0)` stays deterministic, and the leftover `gst_state` column data is simply never read. No reconciliation needed.

---

PR #4521, opened 2026-07-01
fix: correct classification of credit note of b2cl invoice

ljain112:


ljain112:
in favour of: https://github.com/resilient-tech/india-compliance/pull/4522

PR #4545, opened 2026-07-06
fix: Add request timeout handling and auto-retry logic for e-Invoice/e-Waybill

vorasmit:
## What

- Govt servers hang. Requests now time out instead of holding workers.
- After a timeout, next requests to that portal fail fast for 2 mins.
- During an outage, invoices are marked Auto-Retry instead of queueing.

## How

- e-Invoice / e-Waybill: 10s to connect, 30s to respond.
- Other APIs: 10s to connect, no response limit (downloads are slow).
- Timeouts and dropped connections now raise known errors, so auto-retry kicks in.
- Outage is remembered per portal. e-Invoice can be up while e-Waybill is down.
- Only timeouts hold back later requests. Other errors fail on their own.
- Applies to every request, manual ones too. `force` flag is gone.
- The 5 min retry run clears the flag and probes the server again.

## Tests

- Timeout, connect timeout, connection drop, fail fast, per portal, outage reset.

## Note

- e-Waybill error 604 fix is dropped from this PR. Separate PR later, if needed.


india_compliance/gst_india/api_classes/base.py
```diff
@@ -20,11 +20,21 @@
 BASE_URL = "https://asp.resilient.tech"
 DEFAULT_SUPPORT_EMAIL = "api-support@indiacompliance.app"
 
+SERVER_DOWN_CACHE_KEY = "gst_server_down"
```
vorasmit: seperate keys for e-Invoice and e-Waybill portal. possible one is working and other has stopped.

india_compliance/gst_india/api_classes/base.py
```diff
@@ -270,6 +290,26 @@ def handle_server_error(self, error_messages):
...
+            exc=GSPServerError,
+            title=GSPServerError.title,
+        )
+
+    def mark_server_down(self, error):
+        if not self.FAIL_FAST_IF_SERVER_DOWN:
+            return
+
+        # limit exceeded is not an outage
+        if not isinstance(error, GSPServerError) or isinstance(error, GSPLimitExceededError):
```
vorasmit: check instead for timeout instead of gsplimitexceeded

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -605,6 +606,9 @@ def retry_e_invoice_e_waybill_generation():
 
     settings.db_set("is_retry_einv_ewb_generation_pending", 0, update_modified=False)
 
+    # this run is the probe, let requests through
+    frappe.cache.delete_value(SERVER_DOWN_CACHE_KEY)
```
vorasmit: refer to generate_e_invoice / generate_e_waybill flags. we want to keep consistent ux and remove all the checks for is_retry_einv_ewb_generation_pending or even keep force flag, as they both don't add much value. even in sales_invoice, checks can be simpler before enqueue -> only check the flags + setting enabled or disabled.

india_compliance/gst_india/api_classes/base.py
vorasmit: Done. Key is now per portal — `gst_server_down:e-Invoice` / `gst_server_down:e-Waybill`, derived from `API_NAME`.

The 5 min retry run clears both with `delete_keys`. Test covers that one portal going down leaves the other alone.

---

india_compliance/gst_india/api_classes/base.py
vorasmit: Done. Only `GatewayTimeoutError` marks the server down now — GSP down responses, rate limits and connection drops fail on their own, so no reason to hold back later requests.

Note `ConnectTimeout` subclasses both `Timeout` and `ConnectionError`, and `Timeout` is caught first, so a connect hang still counts. Added a test to keep that order.

---

india_compliance/gst_india/utils/e_invoice.py
vorasmit: Done.

- `force` dropped from `generate_e_invoice(s)` / `generate_e_waybill(s)` and from the two JS buttons that passed it.
- `is_retry_einv_ewb_generation_pending` checks removed from both generation paths. The flag now only gates the 5 min retry run, nothing else.
- On submit: check the outage flag, then call `handle_server_errors` instead of setting the status by hand. Same message and status as a failed request, so the user always knows why it says Auto-Retry. Sandbox is handled inside the handler, so that check is gone from here too.

---

india_compliance/gst_india/api_classes/base.py
vorasmit: Correction to the above, as discussed: any `GSPServerError` marks the server down, not just timeouts. Since `GatewayTimeoutError` subclasses it, one check covers both.

Note this also covers `GSPLimitExceededError` (GEN5005). Backing off on a rate limit seems right, but say the word if you want it left out.

---

PR #4546, opened 2026-07-06
fix: Defer e-Invoice/e-Waybill portal cancellation to after-commit job

vorasmit:
## Summary
Refactored the e-Invoice and e-Waybill cancellation flow to defer irreversible portal operations to an after-commit job. This ensures that if a Sales Invoice cancellation rolls back, the portal state remains untouched and doesn't diverge from the database state.

## Key Changes

- **Deferred portal cancellation**: Moved `auto_cancel_e_invoice` and `auto_cancel_e_waybill` calls from synchronous `cancel_e_waybill_e_invoice` hook to a new `cancel_e_invoice_e_waybill_after_commit` function enqueued with `enqueue_after_commit=True`

- **Separated concerns**: 
  - `cancel_e_waybill_e_invoice` now only validates prerequisites and enqueues the job
  - `cancel_e_invoice_e_waybill_after_commit` handles the actual portal cancellation with full gating logic
  - `_cancel_e_invoice` no longer calls `doc.cancel()` (portal-only operation)

- **Updated UI flow**: Modified `e_invoice_actions.js` to clarify that the cancel dialog performs portal cancellation only; the SI is cancelled by a separate request

- **Enhanced test coverage**: 
  - Updated `test_auto_cancel_e_waybill_on_si_cancel` to explicitly test the deferred job execution
  - Added `test_portal_cancel_not_triggered_when_si_cancel_rolls_back` to verify portal remains untouched on cancellation failure
  - Updated `test_cancel_e_invoice` to verify two-step flow: portal cancel first, then SI cancel

## Implementation Details

- Portal cancellation is now idempotent and safe to retry
- If SI cancellation fails/rolls back before commit, the enqueued job is discarded automatically
- Settings validation and auto-cancel gating now live in the deferred job, not the hook
- Maintains backward compatibility with existing e-Invoice/e-Waybill cancellation logic

https://claude.ai/code/session_01VQModrsTicyLZjFRMp54UG

### DIFF
--- a/india_compliance/gst_india/utils/e_invoice.py
+++ b/india_compliance/gst_india/utils/e_invoice.py
@@ -413,7 +413,6 @@ def log_and_process_e_invoice_generation(doc, result, sandbox_mode=False, messag
 
 @frappe.whitelist()
 def cancel_e_invoice(docname: str, values: str | dict | frappe._dict):
-    """Cancel e-Waybill + IRN on the portal only; the SI is cancelled by a separate request."""
     doc = load_doc("Sales Invoice", docname, "cancel")
     values = frappe.parse_json(values)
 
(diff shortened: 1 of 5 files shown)


india_compliance/gst_india/utils/e_invoice.py
vorasmit, line 440: handle case where cancel_irn raises error, but e_waybill successfully cancelled.

PR #4548, opened 2026-07-06
fix: indent JSON data for report JSON downloads

### LINKED ISSUE #4496: UX: for `india_compliance.trigger_file_download` use indentation for json data

vorasmit:
> Explain the details for making this change. What existing problem does the pull request solve?

The **GST Job Work Stock Movement** and **GSTR-1** JSON exports were downloaded as single-line JSON, which is hard to read. This indents those two downloads (2 spaces) at the call site, consistent with the other JSON downloads (HSN summary already indents; e-Waybill is indented server-side).

Per review feedback, `trigger_file_download` itself is left unchanged — each caller decides whether/how to indent based on its data.

> Screenshots/GIFs

N/A — the downloaded `.json` file is now indented instead of single-line.

closes #4496

### DIFF
--- a/india_compliance/gst_india/doctype/gstr_1/gstr_1.js
+++ b/india_compliance/gst_india/doctype/gstr_1/gstr_1.js
@@ -2210,7 +2210,7 @@ class FiledTab extends GSTR1_TabManager {
                     },
                     callback: (r) => {
                         india_compliance.trigger_file_download(
-                            JSON.stringify(r.message.data),
+                            JSON.stringify(r.message.data, null, 2),
                             r.message.filename,
                         );
                         dialog.hide();
--- a/india_compliance/gst_india/report/gst_job_work_stock_movement/gst_job_work_stock_movement.js
+++ b/india_compliance/gst_india/report/gst_job_work_stock_movement/gst_job_work_stock_movement.js
@@ -72,7 +72,7 @@ frappe.query_reports["GST Job Work Stock Movement"] = {
 
     onload: function (query_report) {
         const handle_download = (response) => {
-            india_compliance.trigger_file_download(JSON.stringify(response.data), response.filename);
+            india_compliance.trigger_file_download(JSON.stringify(response.data, null, 2), response.filename);
         };
 
         query_report.page.add_inner_button(__("Export JSON"), function () {


Abdeali099:
@claude 

instead can we check like `if (file_name.endswith(".json"))` and in else do for other file type and keep indent 2 instead of 4.

- also remove the extra comment of `// Pretty-print JSON data...`

Also from `india_compliance/gst_india/report/hsn_wise_summary_of_outward_supplies/hsn_wise_summary_of_outward_supplies.js` from 81 line remove indentation logic

vorasmit:
Done, pushed:

- `trigger_file_download`: flipped to `if (file_name.endsWith(".json")) { … } else { … }`, with the octet-stream type set in the `else`.
- JSON indent is now **2** instead of 4.
- Removed the `// Pretty-print JSON data…` comment.
- `hsn_wise_summary_of_outward_supplies.js`: dropped the `, null, 2` so the caller just passes `JSON.stringify(r.message.data)` and the util handles indentation.

Note: I kept the single comment inside the empty `catch {}` — eslint's `no-empty` needs the block to be non-empty.

---

Abdeali099:
@claude it is not good.

Because we have to do 2 time JSON.stringify whereever this utility being used.

It should be consistent at all place.

vorasmit:
Agreed — the util should serialize once, and every caller should just hand it the data (no caller-side `JSON.stringify`).

Proposed plan:
- `trigger_file_download(data, file_name)`: for `.json`, do `JSON.stringify(data, null, 2)` **once**; write other types as-is.
- Report callers (`hsn_wise_summary_of_outward_supplies`, `gst_job_work_stock_movement`, `gstr_1`): drop `JSON.stringify(...)` and pass the raw object.
- e-Waybill: both `e_waybill_actions.js` and `sales_invoice_list.js` get their payload from the server method `generate_e_waybill_json`, which currently returns `frappe.as_json(ewb_data, indent=4)` (a **string**). To make it consistent I'd have it return the `ewb_data` **dict**, so the client passes an object like everywhere else (util formats it). Downloaded JSON stays identical — just indented by the util.

The only non-trivial part is that last change (a return-type change in the e-Waybill flow; I'd also update `test_e_waybill` if it asserts on the string). Shall I include the e-Waybill server change in this PR, or keep the server returning the string and unify only the report callers?

---

Abdeali099:
@

> Agreed — the util should serialize once, and every caller should just hand it the data (no caller-side `JSON.stringify`).
> 
> Proposed plan:
> 
>     * `trigger_file_download(data, file_name)`: for `.json`, do `JSON.stringify(data, null, 2)` **once**; write other types as-is.
> 
>     * Report callers (`hsn_wise_summary_of_outward_supplies`, `gst_job_work_stock_movement`, `gstr_1`): drop `JSON.stringify(...)` and pass the raw object.
> 
>     * e-Waybill: both `e_waybill_actions.js` and `sales_invoice_list.js` get their payload from the server method `generate_e_waybill_json`, which currently returns `frappe.as_json(ewb_data, indent=4)` (a **string**). To make it consistent I'd have it return the `ewb_data` **dict**, so the client passes an object like everywhere else (util formats it). Downloaded JSON stays identical — just indented by the util.
> 
> 
> The only non-trivial part is that last change (a return-type change in the e-Waybill flow; I'd also update `test_e_waybill` if it asserts on the string). Shall I include the e-Waybill server change in this PR, or keep the server returning the string and unify only the report callers?
> 
> _Generated by [Claude Code](https://claude.ai/code)_

@claude I would suggest do not update the utility bu update the all callers depend on their type, if indent require or not.

- DO not add comments

Abdeali099 (approved):

Issue #4348, opened 2026-06-01
UX: Do Not Close the Quick Entry Dialog When Address Details Are Invalid


Abdeali099:
## Issue

* When a user enters Address details through the Quick Entry dialog and accidentally provides a GSTIN that does not match the selected State, a validation error is thrown.
* The dialog closes immediately after the error, causing all entered data to be lost.
* The user must reopen the dialog and re-enter all the information.

## Expected Behavior

* The Quick Entry dialog should remain open when validation fails.
* Validation errors should be displayed without clearing the entered data, allowing the user to correct the invalid fields and try again.

PR #4551, opened 2026-07-06
feat: auto-select State from GSTIN in quick entry dialog

### LINKED ISSUE #4349: UX: Auto-Select State from GSTIN During Address Creation

vorasmit:
no-docs — minor UX enhancement that auto-fills an existing field; no dedicated documentation page seems necessary. Happy to add docs or reclassify if you'd prefer.

> Explain the details for making this change. What existing problem does the pull request solve?

When creating an Address (or a Customer/Supplier with address details) through the Quick Entry dialog, the **State** had to be selected manually even though the GSTIN already encodes it (the first two digits are the state code). Picking the wrong State triggers the server-side "First 2 digits of GSTIN should match with State Number" validation.

This PR auto-selects the State from the entered GSTIN's state code:
- `boot.py`: expose the state-name → state-number map as `frappe.boot.gst_state_numbers`.
- `utils.js`: add `india_compliance.get_state_from_gstin(gstin)` helper.
- `quick_entry.js`: on GSTIN change, set the State field from the derived state code (works with or without the autofill API).

This reduces manual entry and prevents the GSTIN/State mismatch error (related to #4348).

> Screenshots/GIFs

Entering a GSTIN starting with `27` now sets State to **Maharashtra** automatically.

closes #4349

### DIFF
--- a/india_compliance/boot.py
+++ b/india_compliance/boot.py
@@ -30,6 +30,7 @@ def set_bootinfo(bootinfo):
 
     bootinfo["gst_settings"] = gst_settings
     bootinfo["india_state_options"] = list(INDIAN_STATES)
+    bootinfo["gst_state_by_number"] = {number: state for state, number in INDIAN_STATES.items()}
     bootinfo["ic_api_enabled_from_conf"] = bool(frappe.conf.ic_api_secret)
 
     set_indian_registered_companies(bootinfo)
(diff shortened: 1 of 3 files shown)


Abdeali099:
@claude in boot can't we map number -> state instead of state -> number so when getting state from number it is easy?

vorasmit:
Done — flipped the boot map to `gst_state_by_number` (`{number: state}`), so `get_state_from_gstin` is now a direct lookup (`gst_state_by_number[state_number]`) instead of iterating with `.find()`.

---

Abdeali099 (approved):

PR #4569, opened 2026-07-10
fix: download all categories for gstr 2a

Priyal208:
## Purpose
GSTR-2A now always downloads every section (no per-category selection), and adds the previously-missing ECOM, ECOMA, TDS, and TCS sections so that e-commerce supplies and TDS/TCS credits are captured in GST Inward Supply.

### What changed 

- Added ECOM, ECOMA, TDS, TCS to GSTRCategory / GSTR_2A_ACTIONS and added their parsers in gstr_2a.py (verified against the GSP schema).
- Added the four values to the GST Inward Supply classification options.
- Excluded these categories from reconciliation (inward-supply query) and from amendment-linking (before_save) — they're stored for reference/export but never matched.
- TDS/TCS flat records (no bill no/date) are keyed by return period so a later period can't overwrite an earlier one.
Always download all sections

- Removed the category-selection UI from the download dialog; 2A and 2B now share the same period-based download flow.
- Removed the per-category reconcile_for_* toggles from GST Settings and the auto-reconciliation category filter.
- Removed the set_category_preference endpoint and the stored user preference.
Migration & cleanup

- New patch v16.remove_gstr_2a_category_preferences deletes the obsolete reconcile_for_* Single fields and the purchase_reco_2a_categories user default.
- Dropped the removed defaults from setup and the auto-reconciliation patch.

### Tests

Added parser tests + fixtures for ECOM/ECOMA/TDS/TCS, plus cross-period uniqueness and the amendment-skip guard.
Removed the now-obsolete TestCategoryPreference suite.

### DIFF
--- a/india_compliance/gst_india/doctype/gst_inward_supply/gst_inward_supply.py
+++ b/india_compliance/gst_india/doctype/gst_inward_supply/gst_inward_supply.py
@@ -22,7 +22,11 @@ class GSTInwardSupply(Document):
             self.ims_action = self.previous_ims_action
 
         if self.match_status != "Amended" and (self.other_return_period or self.is_amended):
-            update_docs_for_amendment(self)
+            from india_compliance.gst_india.utils.gstr_2 import NON_RECONCILE_CATEGORY
+
+            # download-only categories are never reconciled: skip amendment linking
+            if self.classification not in NON_RECONCILE_CATEGORY:
+                update_docs_for_amendment(self)
 
     def on_trash(self):
         if self.link_doctype and self.link_name:
@@ -41,6 +45,11 @@ def create_inward_supply(transaction):
         "classification": transaction.classification,
         "supplier_gstin": transaction.supplier_gstin,
     }
+
+    # flat records (TDS/TCS) have no bill no/date; key them by period so a later
+    # period doesn't overwrite an earlier one
+    if not transaction.bill_no:
+        filters["sup_return_period"] = transaction.sup_return_period
 
     if name := frappe.get_value("GST Inward Supply", filters):
         gst_inward_supply = frappe.get_doc("GST Inward Supply", name)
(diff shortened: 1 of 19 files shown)


india_compliance/gst_india/utils/gstr_2/gstr_2a.py
```diff
@@ -250,3 +250,77 @@ def get_invoice_details(self, invoice):
...
+            "sup_return_period": self.return_period,
+            "taxable_value": invoice.amt_ded,
+            "igst": invoice.iamt,
+            "cgst": invoice.camt,
+            "sgst": invoice.samt,
+            "document_value": invoice.amt_ded,
+        }
+
+    def get_transaction_items(self, invoice):
+        pass
```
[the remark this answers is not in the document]
Priyal208: Intentional — TDS/TCS have no supplier profile (no dtcancel), so cancellation can't be captured here; and update_gstins() only bumps GSTIN rows that already exist (never creates them), so for these counterparties it'd be a no-op. Consistent with IMPG, which skips GSTIN tracking for the same reason.

Priyal208:
@greptileai

india_compliance/gst_india/doctype/gst_inward_supply/gst_inward_supply.py
```diff
@@ -22,7 +22,11 @@ def before_save(self):
             self.ims_action = self.previous_ims_action
 
         if self.match_status != "Amended" and (self.other_return_period or self.is_amended):
-            update_docs_for_amendment(self)
+            from india_compliance.gst_india.utils.gstr_2 import NON_RECONCILE_CATEGORY
+
+            # download-only categories are never reconciled: skip amendment linking
+            if self.classification not in NON_RECONCILE_CATEGORY:
+                update_docs_for_amendment(self)
```
vorasmit: check if this works out for ecoma, and then these conditions may not be required

india_compliance/gst_india/utils/gstr_2/gstr_2a.py
```diff
@@ -250,3 +250,77 @@ def get_invoice_details(self, invoice):
             }
         )
         return invoice_details
+
+
+# E-commerce operator supplies u/s 9(5). Same structure as B2B/B2BA.
+class GSTR2aECOM(GSTR2aB2B):
```
vorasmit, line 256: Do we need this at all?

vorasmit (approved):

PR #4620, opened 2026-07-24
fix: address title handling in Address QuickEntry

ljain112:
Issue:

- Address title is not mandatory if the linked document is set.
- Address title was always mandatory due to Frappe field conditions.
- Also, the address title is set based on the GSTIN name.

image
image

### DIFF
--- a/india_compliance/public/js/quick_entry.js
+++ b/india_compliance/public/js/quick_entry.js
@@ -249,7 +249,7 @@ class AddressQuickEntryForm extends GSTQuickEntryForm {
                 fieldname: "_address_title",
                 fieldtype: "Data",
                 default: this.doc.address_title,
-                mandatory_depends_on: "eval: !doc.links?.length && !doc.link_name",
+                mandatory_depends_on: "eval: !doc.link_name",
             },
         ];
     }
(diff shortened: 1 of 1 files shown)


ljain112:
@greptile why 4/5 ?

india_compliance/public/js/quick_entry.js
```diff
@@ -242,14 +242,27 @@ class AddressQuickEntryForm extends GSTQuickEntryForm {
...
+                mandatory_depends_on: "eval: !doc.link_name",
+            },
+        ];
+    }
+
     async render_dialog() {
         const address_fields = this.get_address_fields();
         const fields_to_exclude = address_fields.map(({ fieldname }) => fieldname);
-        fields_to_exclude.push("pincode", "address_line1");
+        fields_to_exclude.push("pincode", "address_line1", "address_title");
```
Abdeali099, line 260: Why exclude?

Abdeali099:
Other LGTM

india_compliance/public/js/quick_entry.js
ljain112: Fields which we are overriding and do not want as standard fields.

PR #4621, opened 2026-07-24
fix: multiple fixes for e-Invoice e-Waybill workflows

vorasmit:
## What

**Background generation (auto / bulk / API)**
- User gets an alert + the doc refreshes when a background e-Invoice / e-Waybill job finishes or fails (before: silent).
- Failure reason is left as a note on the invoice. Unexpected errors also link an Error Log. Bulk failures too.
- Submitting a not-applicable invoice marks it "Not Applicable" and says why (before: stayed Pending).
- Generate / cancel from the desk runs right after save → near instant. Auto / bulk / API still runs in the background.
- Portal calls happen only after the submit / cancel is saved → a rolled-back save never touches the portal.

**Cancelling an invoice that has an IRN / e-Waybill**
- On Cancel: a dialog to cancel the IRN / e-Waybill first, or cancel the invoice only.
- Same dialog for workflow cancel actions (before: skipped).
- If the portal cancel fails, the dialog stays open → retry, cancel invoice only, or close. Invoice cancel is never blocked (before: blocked).
- Auto-cancel setting on → no dialog, portal cancel runs after the invoice cancel is saved (same as before).
- Cancelled invoice with a live IRN / e-Waybill shows a red banner with a Cancel link (open for 24h).
- "Cancel IRN" button no longer cancels the invoice with it.

**Other**
- Transporter / vehicle details are kept even if e-Waybill generation fails (before: lost).
- Desk-triggered portal calls time out at 60s instead of hanging.
- Translations fixed for the new messages.

## Note
- `generate_e_invoice(throw=False)` over API no longer returns 500 on unexpected errors → logs, sets Failed, returns the message. Test updated.
- Other users viewing the same doc still refresh via the usual reload.

https://github.com/user-attachments/assets/431dc172-e272-478a-ad88-74e5fab385e4


india_compliance/gst_india/client_scripts/sales_invoice.js
```diff
@@ -25,9 +25,30 @@ frappe.ui.form.on(DOCTYPE, {
     },
 
     before_submit(frm) {
+        // desk tag: generate now (on_submit), don't queue it
         frm.doc._submitted_from_ui = 1;
     },
 
+    before_cancel(frm) {
+        // desk tag: cancel now (after_cancel), don't queue it
+        frm.doc._cancelled_from_ui = 1;
```
ljain112: flag never recheds backend

ljain112:
Frappe now allows enqueuing the submission; we have to handle that also.

india_compliance/gst_india/client_scripts/sales_invoice.js
ljain112: https://github.com/frappe/frappe/blob/63453d00bae188915874bf662918d1245a875013/frappe/public/js/frappe/form/save.js#L48-L73

india_compliance/gst_india/utils/__init__.py
```diff
@@ -96,10 +97,50 @@ def update_onload(doc, key, value):
...
+        except Exception:
+            # must not raise: frappe rolls back to a savepoint its own commit already dropped
+            frappe.log_error(
+                title=_("Failed to run {0} after response").format(method),
+                message=frappe.get_traceback(),
+            )
+        finally:
+            frappe.flags.in_after_response = False
+
+    commit_after_response(run)
```
vorasmit: after commit

india_compliance/gst_india/client_scripts/e_invoice_actions.js
```diff
@@ -137,37 +121,47 @@ function is_irn_cancellable(frm) {
...
 
     india_compliance.primary_to_danger_btn(d);
     d.show();
 
-    d.show_message(
-        __("Sales invoice will be cancelled along with the IRN."),
-        "yellow",
-        1, // permanent
-    );
+    if (frm.doc.docstatus === 1)
```
sagarvora: think about this

india_compliance/public/js/utils.js
```diff
@@ -580,6 +580,16 @@ Object.assign(india_compliance, {
         return { options, current_year };
     },
 
+    show_cancel_headline(frm, message, on_click) {
+        frm.dashboard
+            .set_headline_alert(`${message} <a href="#">${__("Cancel it")}</a>`, "red", true)
```
sagarvora:
```suggestion
            .set_headline_alert(`${message} <a href="#">${__("Cancel")}</a>`, "red", true)
```

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -94,20 +96,17 @@ def generate_e_invoices(docnames, force=False):
...
-            title=_("e-Invoice generation failed for Sales Invoice {0}").format(docname),
-            message=frappe.get_traceback(),
-            reference_doctype="Sales Invoice",
-            reference_name=docname,
-        )
+    frappe.flags.in_bulk_generation = True
 
     for docname in docnames:
         try:
             generate_e_invoice(docname, force=force)
```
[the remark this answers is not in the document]
vorasmit: It's usually not possible to generate independently as well as in bulk in the same frappe request. Flags go away at the end of request lifecycle.

india_compliance/gst_india/utils/e_invoice.py
```diff
@@ -436,9 +394,7 @@ def _cancel_e_invoice(doc, values):
 
     result = EInvoiceAPI.create(doc).cancel_irn(data)
```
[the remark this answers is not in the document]
vorasmit: We are committing as e-Waybill cancellation is not recoverable; as it's cancelled on the portal.

PR #4629, opened 2026-07-24
refactor: add --skip-assets option to ci for test

Abdeali099:
- Related PR: https://github.com/resilient-tech/india-compliance/pull/4603

image

---

- CI installs `india_compliance` from a local path (`bench get-app <local-folder>`), which triggers bench's cached-app path and **skips `yarn install`**.
- `vue-router`/`vuex` were recently dropped from frappe's own `node_modules` and are now declared in our `package.json`, so they were never installed in CI.
- This caused the esbuild step for the `india_compliance_account` Vue app to fail during CI setup.
- Fix: add `--skip-assets` to the `bench get-app` step — server (Python) tests don't need the JS bundle.
- Regular users are unaffected — installing from a Git URL runs `yarn install` normally.


vorasmit:
Issue was temporary while frappe had removed the vuex etc from it's dependencies.

PR #4661, opened 2026-07-28
Fix e-Invoice print log fallback

vedhamaya:
## Summary
- update the e-Invoice print format to resolve the e-Invoice Log by doc.irn when available
- fall back to the latest active e-Invoice Log linked by sales_invoice when doc.irn is blank or stale
- keep the existing no-preview message when no active e-Invoice Log is available

## Root Cause
Sales Invoice print preview could show no IRN details when the print format relied only on doc.irn, even though an active e-Invoice Log existed for the Sales Invoice.

## Validation
- Cherry-picked the code-only fix from the closed stable-branch PR onto upstream/develop.
- Ran git diff --check upstream/develop..HEAD.


vedhamaya:
Closing this PR as the fix will be handled separately.

PR #4671, opened 2026-07-30
fix: replace search_tds_sections with get_tds_section_options for improved TDS section retrieval

ljain112:
reafactor for simplicity

### DIFF
--- a/india_compliance/income_tax_india/constants/custom_fields.py
+++ b/india_compliance/income_tax_india/constants/custom_fields.py
@@ -1,7 +1,25 @@
-from india_compliance.income_tax_india.constants import OLD_TDS_SECTIONS, TDS_ENTITY_TYPE
+import frappe
+
+from india_compliance.income_tax_india.constants import (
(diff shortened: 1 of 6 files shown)


india_compliance/income_tax_india/constants/custom_fields.py
```diff
@@ -1,8 +1,26 @@
...
+)
 
 old_tds_section_options = "\n" + "\n".join(sorted(OLD_TDS_SECTIONS))
 tds_entity_type_options = "\n" + "\n".join(sorted(TDS_ENTITY_TYPE))
 
+tds_section_options = frappe.as_json(
+    [
+        {
+            "label": get_tds_section_value(section),
+            "value": get_tds_section_value(section),
```
Abdeali099, line 17: Instead, can these be moved to `income_tax_india/constants` itself? That way, we don't need to import constants from another module, and everything stays better organized.

Abdeali099:
@ljain112 LGTM

PR #4674, opened 2026-07-30
fix: update cancel_depreciation_entries logic and add test cases for asset depreciation

### LINKED ISSUE #4640: Depreciation Entries Skips Remaining Finance Books on Empty Row
### Information about bug

**Related issue:** #4638, PR #4639

### Description

`cancel_depreciation_entries()` uses `return` instead of `continue` when iterating over an asset's finance books and encounters a row with no `finance_book` set. This causes the function to exit entirely, skipping all subsequent finance book rows — including those marked `for_income_tax` that should have their depreciation entries cancelled.

### Impact

When an asset has multiple finance books and the first row has no finance book assigned (the default schedule), any income-tax finance book rows that follow are never processed. As a result, depreciation journal entries for the income-tax book are left submitted during asset sale/scrap, leading to incorrect financial statements for the year of sale.

### Root Cause

```python
# Before (buggy):
for row in asset_doc.get("finance_books"):
    if not row.finance_book:
        return          # ← exits function, discarding remaining rows

# After (fixed):
for row in asset_doc.get("finance_books"):
    if not row.finance_book:
        continue        # ← skips to next row
```

The intent of the check is to skip rows that represent the default (non-finance-book) depreciation schedule. `return` was likely a copy-paste error — adjacent code (the `cancel_depreciation_entries` method on the Asset Depreciation Schedule doctype, line 107 in the same code path) also uses `continue`.

### Trigger Condition

The bug triggers when **all** of these are true:
1. Asset has multiple finance book rows in `finance_books` child table
2. The first row (by ordering in the table) has an empty `finance_book` field
3. A subsequent row references a Finance Book marked `for_income_tax = 1`
4. `cancel_depreciation_entries()` is called (e.g., during asset sale in current fiscal year)

### Reproduction

1. Create an Asset with two finance book rows:
   - Row 1: `finance_book` = empty (default depreciation schedule)
   - Row 2: `finance_book` = "IT Act Book" (with `for_income_tax = 1`)
2. Submit the asset (generates depreciation schedules)
3. Manually create and link Journal Entries to the IT Act depreciation schedule rows
4. Call `cancel_depreciation_entries(asset, date_in_current_fy)`
5. **Observed:** IT Act JE remains submitted (docstatus = 1)  
   **Expected:** IT Act JE is cancelled (docstatus = 2)

### Fix Applied

- `return` → `continue` on line 129
- `fb_for_income_tax_map[key]` → `.get(key)` on line 131 as a defensive guard against missing keys

### Regression Test

`test_asset_depreciation_schedule.py::TestAssetDepreciationByIncomeTaxAct::test_cancel_first_row_no_fb_processes_it_row`

### Version

erpnext: 16.29.0 (HEAD)
india_compliance: 16.7.0 (HEAD)

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ljain112:
closes: https://github.com/resilient-tech/india-compliance/issues/4640
Issue only if the user has upgraded from v14.
The latest version of erpnext already has validation for a mandatory finance book for more than one finance book.

### DIFF
--- a/india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
+++ b/india_compliance/income_tax_india/overrides/asset_depreciation_schedule.py
@@ -126,7 +126,7 @@ def cancel_depreciation_entries(asset_doc, date):
 
     for row in asset_doc.get("finance_books"):
         if not row.finance_book:
-            return
+            continue
 
         if not fb_for_income_tax_map[row.finance_book]:
             continue
--- a/india_compliance/income_tax_india/overrides/test_asset_depreciation_schedule.py
+++ b/india_compliance/income_tax_india/overrides/test_asset_depreciation_schedule.py
@@ -257,38 +257,51 @@ class TestAssetDepreciationByIncomeTaxAct(IntegrationTestCase):
             flt(12750 * 366 / 365, 2),
         )
 
-    def test_monthly_depreciation_totals_the_annual_rate(self):
+    def test_second_year_applies_full_rate_after_a_halved_first_year(self):
         asset = self._create_asset(
-            "2024-04-01",
-            depreciation_start_date="2024-04-30",
+            "2024-11-01",
+            depreciation_start_date="2025-03-31",
+            finance_book=self.fb_income_tax.name,
+        )
+
+        # the proviso halves only the first year; the second is 15% of 92500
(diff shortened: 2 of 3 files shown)


ljain112:
No major fix, just AI-generated test cases.

PR #4723, opened 2026-08-10
fix: convert HSN field in transactions to `Autocomplete`

### LINKED ISSUE #4717: HSN/SAC Field Type | Data to Link
### Information about bug

Currently multiple doctypes in ERPNext, Supplier Quotation in current example, have HSN/SAC field as a Data field.
However, having the field type as 'Link' would help a lot where the user would be able to pick the HSN/SAC based on the HSN SAC Master List instead of manually entering it initially or changing it on transactions.

### Version

Frappe : 15.116.0
ERPNext: 15.118.1
India Compliance: 15.30.2

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ljain112:
Closes: https://github.com/resilient-tech/india-compliance/issues/4717

Existing behaviour to allow any HSN is also supported.

image

### DIFF
--- a/india_compliance/gst_india/constants/custom_fields.py
+++ b/india_compliance/gst_india/constants/custom_fields.py
@@ -685,7 +685,7 @@ CUSTOM_FIELDS = {
         {
             "fieldname": "gst_hsn_code",
             "label": "HSN/SAC",
-            "fieldtype": "Data",
+            "fieldtype": "Autocomplete",
             "fetch_from": "item_code.gst_hsn_code",
             "insert_after": "description",
             "allow_on_submit": 1,
@@ -807,7 +807,7 @@ CUSTOM_FIELDS = {
         {
             "fieldname": "gst_hsn_code",
             "label": "HSN/SAC",
-            "fieldtype": "Data",
+            "fieldtype": "Autocomplete",
             "fetch_from": "item_code.gst_hsn_code",
             "insert_after": "description",
             "allow_on_submit": 1,
--- a/india_compliance/gst_india/constants/custom_fields.py
+++ b/india_compliance/gst_india/constants/custom_fields.py
@@ -47,6 +47,18 @@ party_fields = [
     },
 ]
 
+transaction_hsn_code_field = {
+    "fieldname": "gst_hsn_code",
+    "label": "HSN/SAC",
+    "fieldtype": "Autocomplete",
+    "fetch_from": "item_code.gst_hsn_code",
+    "insert_after": "description",
+    "allow_on_submit": 1,
+    "print_hide": 1,
+    "fetch_if_empty": 1,
+    "translatable": 0,
+}
+
 CUSTOM_FIELDS = {
     # Subcontracting: Tax Fields
     "Subcontracting Order": [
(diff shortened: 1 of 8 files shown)


india_compliance/gst_india/constants/custom_fields.py
```diff
@@ -685,7 +685,7 @@
         {
             "fieldname": "gst_hsn_code",
             "label": "HSN/SAC",
-            "fieldtype": "Data",
+            "fieldtype": "Autocomplete",
             "fetch_from": "item_code.gst_hsn_code",
```
Abdeali099: Should we create one constant dictionary and reuse it in both places? We can unpack it and update the required keys as needed.

india_compliance/gst_india/utils/__init__.py
```diff
@@ -499,6 +499,29 @@ def get_hsn_settings():
     return validate_hsn_code, valid_hsn_length
 
 
+@frappe.whitelist()
+def get_hsn_code_list(txt: str | None = None, limit: int = 20):
+    """
+    GST HSN Code is seeded public data, readable by all roles, hence no permission check.
```
Abdeali099: IMO, this should be a comment rather than a docstring.

india_compliance/gst_india/utils/__init__.py
```diff
@@ -499,6 +499,29 @@ def get_hsn_settings():
...
+    query = (
+        frappe.qb.from_(hsn_code)
+        .select(hsn_code.name.as_("value"), hsn_code.name.as_("label"), hsn_code.description)
+        .orderby(hsn_code.name)
+        .limit(cint(limit))
+    )
+
+    validate_hsn_code, valid_hsn_length = get_hsn_settings()
+    if validate_hsn_code and valid_hsn_length:
+        query = query.where(hsn_code.name.like("_" * min(valid_hsn_length) + "%"))
```
Abdeali099: Could we use a length check here instead of `LIKE`? `valid_hsn_length` contains the allowed lengths (e.g. `(6, 8)`), but `min(valid_hsn_length)` with `LIKE '______%'` only checks that the length is **at least 6**, so a 7-digit HSN would also match even though 7 isn't a valid length.

For example, with valid lengths `(6, 8)`:

* `123456` → valid
* `12345678` → valid
* `1234567` → should be invalid, but the current condition matches it.

Could we instead check the actual length against the valid lengths, something like:

```python
query = query.where(Length(hsn_code.name).isin(valid_hsn_length))
```

This would ensure only HSNs with one of the configured valid lengths are matched.


> Need to check PostgreSQL support

india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.js
```diff
@@ -4,6 +4,7 @@
 frappe.ui.form.on("Bill of Entry", {
     setup(frm) {
         india_compliance.setup_itc_claim_period_query(frm);
+        india_compliance.set_hsn_code_autocomplete(frm);
```
Abdeali099, line 7: Why can't we add this directly to `transaction.js`?

india_compliance/public/js/transaction.js
```diff
@@ -211,6 +222,14 @@ function set_gst_tax_breakup_on_load(doctype) {
     });
 }
 
+function set_hsn_code_autocomplete(doctype) {
+    frappe.ui.form.on(doctype, {
```
Abdeali099, line 226: Why do we need a separate function? Can we handle this directly in the loop above?

Abdeali099:
@ljain112, other than that, LGTM.

ljain112:
@vorasmit Is this acceptable?

india_compliance/public/js/transaction.js
ljain112: it does not contain pos and material request doctype

india_compliance/gst_india/doctype/bill_of_entry/bill_of_entry.js
ljain112: India Compliance doctype

ljain112:
@greptileai review

PR #4739, opened 2026-08-11
fix: remove white spaces from hsn description

ljain112:
Issue: Govt rejects whitespace at the end in hsn description

### DIFF
--- a/india_compliance/gst_india/utils/gstr_1/gstr_1_json_map.py
+++ b/india_compliance/gst_india/utils/gstr_1/gstr_1_json_map.py
@@ -14,7 +14,6 @@ from india_compliance.gst_india.utils import (
 )
 from india_compliance.gst_india.utils.gstr_1 import (
     CATEGORY_SUB_CATEGORY_MAPPING,
-    HSN_DESCRIPTION_LIMIT,
     SUB_CATEGORY_GOV_CATEGORY_MAPPING,
     SUBCATEGORIES_NOT_CONSIDERED_IN_TOTAL_TAX,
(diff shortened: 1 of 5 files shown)


ljain112:
@greptileai 

HSN autocomplete — FALSE POSITIVE. The review assumed search_widget's six-positional-arg contract, but gst_hsn_code is fieldtype Autocomplete, not Link. Autocomplete.execute_query_if_exists ([autocomplete.js:223-226](vscode-webview://0uhlqd1vssqhhnlnf58vu6tgqrip5e5dfbbq04qai9n8qrg8le44/home/lj/develop/apps/frappe/frappe/public/js/frappe/form/controls/autocomplete.js#L223-L226)) does frappe.call({ method: args.query, args: { txt } }) — a direct method call with keyword args that never reaches search_widget. The current (txt, limit) signature matches exactly. My initial search_widget reproduction was testing the wrong entry point; you were right to send me back to the query side. No change needed.

ljain112:
@greptileai review

india_compliance/gst_india/utils/gstr_1/gstr_1_json_map.py
```diff
@@ -1261,6 +1262,9 @@ def map_uom(self, uom, data=None):
 
         return f"OTH-{UOM_MAP.get('OTH')}"
 
+    def truncate_description(self, description, data=None):
```
vorasmit, line 1265: simple

ljain112: Value formatter uses two args, so separate  function.

PR #4747, opened 2026-08-13
fix: add multiple doctypes to `audit_trail_doctypes` hook

Anish59312:
As per custom request `Asset Depreciation Schedule` should have track_changes in sync with `Enable Audit Trail` of `Account Settings`. Run `bench migrate` after this change to fix the issue.

Some other DocTypes may have get the same issue. Selected ones are finally added to hooks in this PR.

#### ERPNext
- [ ] ~Process Period Closing Voucher (majorly machine generated)~
- [x] Cost Center Allocation (not direct transaction, but governs GL entry split) *
_____
  `track_changes` true but editable
- [ ] ~Repost Payment Ledger (not set in voucher_no)~
- [x] Exchange Rate Revaluation. (makes JE and not GL)
- [x] Asset Value Adjustment (makes JE and not GL)
- [ ] ~Bank Transaction (no GL but has allow on submit fields)~

#### HRMS

separate PR made in HRMS
- [x] Gratuity *
- [X] Salary Slip
- [X] Payroll Entry
- [X] Expense Claim
____
 `track_changes` true but editable
- [x] Leave Encashment *
- [ ] ~Employee Advance (no direct GL)~
- [ ] ~Full and Final Statement~

### DIFF
--- a/india_compliance/hooks.py
+++ b/india_compliance/hooks.py
@@ -467,10 +467,7 @@ audit_trail_doctypes = [
     # Additional ERPNext DocTypes that constitute "Books of Account"
     "Asset Depreciation Schedule",
     "POS Invoice",
-    # Affects the GL entry creation
     "Cost Center Allocation",
-    # ERPNext DocTypes that have track_changes enabled by default
-    # Need to add here to stop them from being disabled when Audit Trail is enabled
     "Exchange Rate Revaluation",
     "Asset Value Adjustment",
     # India Compliance DocTypes that make GL Entries
--- a/india_compliance/hooks.py
+++ b/india_compliance/hooks.py
@@ -465,6 +465,7 @@ audit_trail_doctypes = [
     "Stock Reconciliation",
     "Subcontracting Receipt",
     # Additional ERPNext DocTypes that constitute "Books of Account"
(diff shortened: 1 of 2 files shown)


Anish59312:
@greptileai

india_compliance/hooks.py
```diff
@@ -465,7 +465,14 @@
     "Stock Reconciliation",
     "Subcontracting Receipt",
     # Additional ERPNext DocTypes that constitute "Books of Account"
+    "Asset Depreciation Schedule",
     "POS Invoice",
+    # Affects the GL entry creation
+    "Cost Center Allocation",
+    # ERPNext DocTypes that have track_changes enabled by default
```
ljain112: remove unwanted comments and patch re-run required

Anish59312: after migration hooks runs this function, so patch changes may not be required

Anish59312: https://github.com/Anish59312/india-compliance/blob/5f9dfb6952d4d0134663b5a8fab0b81c498dfa94/india_compliance/audit_trail/setup.py#L45

ljain112 (approved):

PR #4748, opened 2026-08-13
refactor: GSTR-1 books processing

vorasmit:
Books side of GSTR-1 mapping now lives one file per return section, same as the portal side already does.

## What

- Was: one big class built rows for every section.
- Now: each section owns its own file and builds its own rows.
- `gstr_1_books_map.py` just queries and hands off. ~460 lines lighter.
- Same section is now read and written in one place.

## Note

- Nothing else changes in what gets filed.
- New tests: every section reachable both ways, no invoice type left unclaimed, HSN lookup asks only what it needs.


Priyal208:
@greptileai

PR #4758, opened 2026-08-13
fix: handle missing Email Template reference doctype

Anuradha145:
## What changed

- Check whether the Email Template reference_doctype database column exists before adding it to the reconciliation-template query.
- Preserve the broader query on schemas that provide the column.
- Add regression coverage for Frappe v16 schemas where the column is absent.

## Why

The v16 update_reconciliation_email_template_dates patch currently fails during migrate with MySQL error 1054 when Email Template has no reference_doctype column. This prevents all later migrations from running.

## Validation

- Python bytecode compilation passed.
- git diff --check passed.
- Added a regression test asserting the fallback query.


ljain112:
Thank you for your contribution.

Continued: https://github.com/resilient-tech/india-compliance/pull/4759

PR #4763, opened 2026-08-14
fix: abs before round 

Priyal208:
### What
- Credit notes could send a paisa less to the portal (e-invoice / e-waybill) than books show.
- Cause: amount made positive **after** rounding, should be before.

### Why
- Rounding is not symmetric around zero.
  `-2.675` → round then flip = `2.67`, flip then round = `2.68`.
- Credit notes are negative in books, so such amounts went out one paisa low.

### How
- Take magnitude (or flip sign) first, round after.
- 10 spots: invoice totals, grand total, discount, item value, qty, unit rate, paid amount, outstanding, round-off.

### DIFF
--- a/india_compliance/gst_india/overrides/test_transaction_data.py
+++ b/india_compliance/gst_india/overrides/test_transaction_data.py
@@ -329,6 +329,19 @@ class TestTransactionData(IntegrationTestCase):
             ],
         )
 
+    @change_settings("System Settings", {"rounding_method": "Banker's Rounding (legacy)"})
+    def test_rounded_is_symmetric_about_zero(self):
+        """-x must round to the same size as x."""
+        for value in (1.005, 2.675, 0.125, 1234.565):
+            for precision in (2, 3):
+                self.assertEqual(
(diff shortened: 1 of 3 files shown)


ljain112:
Is this related to : https://github.com/frappe/frappe/pull/39562

ljain112:
Ideally, we should fix the rounded function to always round the absolute value, then return the value the correct value

Priyal208:
> Is this related to : [frappe/frappe#39562](https://github.com/frappe/frappe/pull/39562)

yes.. this bug is using `_bankers_rounding_legacy`.  is it good idea to change a legacy function ( to fix at root )?

ljain112:
> > Is this related to : [frappe/frappe#39562](https://github.com/frappe/frappe/pull/39562)
> 
> yes.. this bug is using `_bankers_rounding_legacy`. is it good idea to change a legacy function ( to fix at root )?

We can fix are rounded function to always use abs value instead of handling it everywhere.

ljain112 (approved):

PR #4770, opened 2026-08-17
fix: reverse tax amount proportionately in advance tax reversal

ljain112:
Issue:

Advance of 500 + 90 GST (CGST 45 + SGST 45), 590 banked. Two Sales Orders with 100 earmarked against each, leaving 300 unallocated. Nothing reversed yet, so the whole 90 is still pending.

Why one payment becomes three rows
get_advance_payment_entries runs two queries and concatenates them with no dedup — and the reconciliation tool calls it with against_all_orders=True:

if order_list or against_all_orders:
    ...   # one row per Payment Entry Reference pointing at an order
if include_unallocated:
    ...   # plus one row for the leftover balance
Every row carries the same reference_name — the Payment Entry name. So one advance arrives as three rows.

Before — the pot added to every row

row          net    + pot    offered
SO-1         100    + 90  =  190
SO-2         100    + 90  =  190
unallocated  300    + 90  =  390
                            ─────
                             770     ← the payment is worth 590
Over-stated by 180 — the pot counted once too often per extra row. That surplus is spendable: allocate_entries caps an allocation at pay["amount"], so it reaches an invoice and the reconcile then fails late with "Outstanding amount … is less than the total allocated amount with taxes".

### DIFF
--- a/india_compliance/gst_india/overrides/payment_entry.py
+++ b/india_compliance/gst_india/overrides/payment_entry.py
@@ -4,6 +4,7 @@ from erpnext.accounts.utils import create_payment_ledger_entry
 from erpnext.controllers.accounts_controller import get_advance_payment_entries
 from frappe import _
 from frappe.contacts.doctype.address.address import get_default_address
+from frappe.model.meta import get_field_precision
 from frappe.query_builder.functions import Sum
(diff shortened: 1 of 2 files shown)


ljain112:
@greptileai review

PR #4781, opened 2026-08-18
fix: handle multi-currency in GST on advance payment

ljain112:
## fix: multi-currency GST advance reversal

GST is always company currency; a Payment Entry's own amounts are in the party account
currency. Every defect below is one used where the other belonged — invisible in an
INR-only company, where the two coincide.

### Root cause

`get_taxes_summary` selected `pe.paid_amount` (transaction) while deriving `taxable_amount`
from it as if it were base. For a 5 USD @ 100 advance that made `paid_proportion` `5/95`
instead of `500/590`, crushing every FX allocation to ~5%. Now selects `base_paid_amount`,
which makes `tax_amount / taxable_amount` a dimensionless rate and fixes three consumers
at once.

### Fixed

- **Reconciliation offered ~16× the advance.** The pending-GST pot (INR) was added straight
  onto `pe.amount` (USD) — `95.00` offered for an advance worth `5.90`. Anything allocated
  above the real balance then failed late with ERPNext's *"Allocated amount cannot be
  greater than unadjusted amount"*.
- **The reversal wrote rupees into a dollar column.** `credit_in_account_currency` got the
  base figure — `18.00` where the USD receivable should move by `0.18`.
- **The over-allocation guard rejected every foreign advance.** It compared base tax against
  a transaction-currency outstanding (`18 + 1 = 19` vs `1.18`). Now compared in company
  currency via ERPNext's own `calculate_base_allocated_amount_for_reference`, so the `> 1`
  slack stays one rupee and no precision is lost dividing the tax down.
- **FX over-allocation never raised in the advances table.** `set_and_validate_advances_with_gst`
  compared a party-currency total against `base_grand_total`. Now mirrors ERPNext's
  `set_advances` (`advances.py:36-39`).
- **GSTR-1 11B and GST Advance Detail reported the wrong currency.** 11B paired 1 USD of
  taxable value against 18 INR of tax, bucketing the row at a 1800% rate; the report showed
  `paid=5.0` beside `gst_paid=90.0` on columns declared as company currency.

Also: the party-side rate now comes from `PaymentEntry.get_exchange_rate()` rather than a
hardcoded `source_exchange_rate`, so `Pay` (RCM) advances take the target rate; amounts round
to field precision instead of a hardcoded `2`; and the rate falls back to 1 for zero as well
as null, since ERPNext accepts a zero exchange rate on save (previously a `ZeroDivisionError`).

### DIFF
--- a/india_compliance/gst_india/overrides/payment_entry.py
+++ b/india_compliance/gst_india/overrides/payment_entry.py
@@ -226,7 +226,21 @@ def _get_gl_for_advance_gst_reversal(payment_entry, reference_row, via_reconcili
         return gl_dicts
 
     outstanding_amount = reference_row.outstanding_amount
-    if via_reconciliation:
+    if outstanding_amount is None:
+        # erpnext stops refreshing the row's reference details once exchange_gain_loss is set on
+        # it, so resolve them the way it would have. Journal Entry has no outstanding_amount
+        # field, hence get_reference_details rather than a direct read. Not reduced by this
+        # allocation yet, so it needs no adding back.
+        from erpnext.accounts.doctype.payment_entry.payment_entry import get_reference_details
+
+        outstanding_amount = get_reference_details(
+            reference_row.reference_doctype,
+            reference_row.reference_name,
+            payment_entry.party_account_currency,
+            payment_entry.party_type,
+            payment_entry.party,
+        ).outstanding_amount
+    elif via_reconciliation:
         # add back allocated_amount to outstanding_amount for comparison
         outstanding_amount += reference_row.allocated_amount
 
--- a/india_compliance/gst_india/overrides/payment_entry.py
+++ b/india_compliance/gst_india/overrides/payment_entry.py
@@ -230,20 +230,17 @@ def _get_gl_for_advance_gst_reversal(payment_entry, reference_row, via_reconcili
         # add back allocated_amount to outstanding_amount for comparison
         outstanding_amount += reference_row.allocated_amount
 
-    base_outstanding_amount = flt(
-        outstanding_amount * exchange_rate, payment_entry.precision("base_paid_amount")
-    )
-    total_allocation = total_amount + payment_entry.calculate_base_allocated_amount_for_reference(
-        reference_row
-    )
-    excess_allocation = total_allocation - base_outstanding_amount
+    total_allocation = total_amount_in_account_currency + reference_row.allocated_amount
+    excess_allocation = total_allocation - outstanding_amount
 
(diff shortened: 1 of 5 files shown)


india_compliance/gst_india/overrides/payment_entry.py
```diff
@@ -453,8 +472,18 @@ def get_taxes_summary(company, payment_entries):
             Sum(gl_entry.credit_in_account_currency).as_("tax_amount"),
             Sum(gl_entry.debit_in_account_currency).as_("tax_amount_reversed"),
             pe.name.as_("payment_entry"),
-            pe.paid_amount,
+            pe.base_paid_amount.as_("paid_amount"),
             pe.unallocated_amount,
+            # ERPNext allows a zero exchange rate to be saved, so guard zero as well as null
+            IfNull(
```
vorasmit: simplify

india_compliance/gst_india/overrides/sales_invoice.py
```diff
@@ -355,10 +355,13 @@ def set_and_validate_advances_with_gst(doc):
...
         allocated_amount_with_taxes += advance.allocated_amount
 
-    excess_allocation = flt(
-        flt(allocated_amount_with_taxes, 2) - (doc.base_rounded_total or doc.base_grand_total),
-        2,
-    )
+    if doc.get("party_account_currency") == doc.company_currency:
+        total = doc.get("base_rounded_total") or doc.base_grand_total
+    else:
+        total = doc.get("rounded_total") or doc.grand_total
```
vorasmit, line 364: simplify?

india_compliance/gst_india/overrides/test_advance_payment_entry.py
```diff
@@ -660,7 +661,103 @@ def make_payment_reconciliation(payment_doc, invoice_doc, amount):
 }
 
 
-class TestPaymentReconciliationMatrix(IntegrationTestCase):
+class LedgerNetMixin:
+    """Net ledger / GSTR-1 advance assertions shared by the single- and multi-currency suites."""
+
+    def _advance_detail_summary(self, payment_doc):
```
vorasmit, line 745: test for different exchange rates

vorasmit:
Later:

- [ ] Refund workflow

india_compliance/gst_india/overrides/sales_invoice.py
ljain112: If the party account is in company currency and the doc currency is different, then we need the base_rounded_total, and if the party account is in a different currency, then we need rounded_total

PR #4800, opened 2026-08-24
fix: show party name instead of party id in purchase reco

ljain112:
Issue: For some suppliers, the party name is shown, and for others, the party code is shown in Purchase Reco.

So the name of the party will be like:

- First, if it's a purchase invoice, the name will be from the supplier_name field in the Purchase Invoice.
- Then it will be guessed based on GSTIN, and party_name (supplier_name) will be shown.
- Otherwise, the name from the GST Inward Supply will be shown.


ljain112:
@greptileai review

ljain112:
@greptileai GSTIN in bill of entry query is just for informational purposes; while reconciling, we are not using it.

ljain112:
@greptileai review

india_compliance/gst_india/doctype/purchase_reconciliation_tool/__init__.py
```diff
@@ -604,6 +604,7 @@ def get_fields(self, additional_fields=None):
             self.BOE.bill_of_entry_date.as_("bill_date"),
             self.BOE.posting_date,
             self.BOE.company_gstin,
+            Max(self.PI.supplier).as_("supplier"),
```
Anish59312, line 607: this does actually set the wrong supplier 
to replicate try to send email to such supplier 
supplier_name can be of supplier-2 and supplier email of supplier-1

ljain112: For now, we are assuming that all purchases will be from a single supplier.

PR #4855, opened 2026-09-04
fix: validate party type in GSTIN Status report

karm1000:


sagarvora:
should be a constant and frozenset / tuple.

PR #4870, opened 2026-09-16
fix: default gst vehicle type in e-way bill

Priyal208:
### What

Generate and Update Vehicle Info dialogs open with the invoice's vehicle type. Before, an ODC invoice showed Regular.

### Why

Frappe fills dialog defaults the same way a user types, so Mode of Transport's default fired the "set vehicle type from mode" step and forced Regular for Road. Easy to send Regular to the portal while changing something else.

### How

For Road, use the invoice's own vehicle type; fall back to Regular only when there is none. Ship still forces ODC, Rail and Air still blank.


india_compliance/gst_india/client_scripts/e_waybill_actions.js
```diff
@@ -1383,8 +1383,11 @@ function are_transport_details_available(doc) {
     );
 }
 
-function update_vehicle_type(dialog) {
-    dialog.set_value("gst_vehicle_type", get_vehicle_type(dialog.get_values(true)));
+function update_vehicle_type(dialog, frm) {
+    const values = dialog.get_values(true);
+
+    const saved_type = values.mode_of_transport == "Road" && frm.doc.gst_vehicle_type;
```
vorasmit: can be part of get-vehicle_type

PR #4880, opened 2026-09-18
fix: sync transport fields after submit

### LINKED ISSUE #4823: Update Transporter Information on Submitted Sales Invoice
### Is your feature request related to a problem? Please describe.

Currently the system has the provision to update the Transporter information on Submitted Sales Invoice via E-Way Bill button -> Update Transporter option.

Image

However, instead of having the transporter information updated via the Button, the system needs to have the Transporter Information fields editable after Submit where the user can modify the values and this syncs with the E-Way Bill Portal as an Update.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

Priyal208:
**What**
- Transporter, GST Transporter ID, vehicle, mode, LR fields stay editable on a submitted invoice after the e-Waybill exists.
- Press Update with such a change, the portal update dialog opens prefilled with it. Vehicle and transporter both changed → one dialog, two pages, vehicle first.
- Fields lock only when the portal cannot take an update: API off, e-Waybill expired, no submit permission. Same rule as the two buttons.
- Driver fields save normally, portal has no such field.

**Why**
- Today these fields grey out the moment an e-Waybill is generated. Only way to change transporter or vehicle is the e-Waybill menu. Users edit the form and expect it to sync.

**How**
- Custom fields lose the "read-only once e-Waybill" rule, patch does the same on existing sites. Distance stays locked, no portal update for it.
- Form's Update is intercepted when portal fields changed; Frappe skips the usual hooks for Update, so save itself is wrapped.
- Old Update Vehicle Info and Update Transporter dialogs merged into one dialog, same fields, same server methods. Buttons use it too.
- Server guard still blocks a raw save of portal fields (REST, scripts), lets driver through.

Fixes #4823


Priyal208:
@greptileai

india_compliance/patches/v16/unlock_transport_fields_after_e_waybill.py
```diff
@@ -0,0 +1,22 @@
+import frappe
```
vorasmit: not needed. keep the default value as "", and recreate custom fields.

india_compliance/gst_india/overrides/transaction.py
```diff
@@ -2024,12 +2025,12 @@ def validate_transporter_fields_after_submit(doc, method=None):
     if not changed_fields:
         return
 
-    if doc.get("ewaybill"):
+    if doc.get("ewaybill") and set(changed_fields) - set(LOCAL_ONLY_TRANSPORTER_FIELDS):
```
vorasmit: simplify. transporter_name need not be allowed on submit.
read_only and allowed on submit

vorasmit: Fields like transporter_name still needs to be allowed on submit, so that it allows to update

